PR #190, opened 2021-02-25
Implement ID in autotune

ShishirH:
Signed-off-by: Shishir Halaharvi <shalahar@redhat.com>

### DIFF
--- a/src/main/java/com/autotune/dependencyAnalyzer/util/Util.java
+++ b/src/main/java/com/autotune/dependencyAnalyzer/util/Util.java
@@ -1,5 +1,23 @@
+/*******************************************************************************
+ * Copyright (c) 2020, 2021 Red Hat, IBM Corporation and others.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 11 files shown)


src/main/java/com/autotune/dependencyAnalyzer/util/Util.java
```diff
@@ -0,0 +1,26 @@
...
+ *******************************************************************************/
+package com.autotune.dependencyAnalyzer.util;
+
+/**
+ * Contains methods that are of general utility in the codebase
+ */
+public class Util
+{
+	public static int generateID(Object object) {
+		return object.toString().hashCode();
```
dinogun, line 24: As mentioned, we need a sha256 id here based on the object string.

ShishirH: Examples of IDs generated by hashcode and sha256:
Hashcode: -987261259
Sha256: 417b34d90526c4ca6e064d5468f94ce79334c5ebbde6eec8e8da486085750e4e

Personally, as the ID will be passed as a parameter, to autotune APIs,  I was thinking the hashcode might be simpler, and it can be quite easy to look at the difference between IDs for different objects.

However, if you prefer sha256, I'll push the change that I've made

dinogun: I'd prefer sha256 here.

ShishirH: @dinogun Pushed the commit changed algorithm to sha256

dinogun (approved):
lgtm

PR #215, opened 2021-03-24
Create Python REST APIs

agarwalrounak:
Implement POST and GET handlers to serve POST and GET request types respectively.

Signed-off-by: Rounak Agarwal rounag16@in.ibm.com


hyperparameter_tuning/service.py
```diff
@@ -0,0 +1,124 @@
...
+    def _set_response(self, status_code, return_value):
+        # TODO: add status_message
+        self.send_response(status_code)
+        self.send_header('Content-type', 'text/html')
+        self.end_headers()
+        self.wfile.write(return_value.encode('utf-8'))
+
+    def do_POST(self):
+        # TODO: perform better regex search
+        if re.search('/experiment_trials', self.path):
```
dinogun, line 43: The URL needs to be defined as a global var and referenced here.

hyperparameter_tuning/service.py
```diff
@@ -0,0 +1,124 @@
...
+                        self._set_response(400, "-1")
+                elif json_object["operation"] == "EXP_TRIAL_GENERATE_SUBSEQUENT":
+                    if json_object["id"] in autotune_object_ids:
+                        get_search_create_study(json_object["id"], json_object["operation"], json_object["url"])
+                        trial_number = optuna_hpo.TrialDetails.trial_number
+                        self._set_response(200, str(trial_number))
+                    else:
+                        self._set_response(400, "-1")
+                elif json_object["operation"] == "EXP_TRIAL_RESULT":
+                    if json_object["id"] in autotune_object_ids and json_object["trial_number"] == optuna_hpo.TrialDetails.trial_number:
```
dinogun, line 66: We cannot be referencing optuna_hpo directly here. We could be using any algo. I would suggest that we have a separate function to get the number to which you pass the id. It should check the algo that is being used for that id and accordingly get the trial_number,

hyperparameter_tuning/service.py
```diff
@@ -0,0 +1,124 @@
...
+                        optuna_hpo.TrialDetails.trial_result = json_object["trial_result"]
+                        optuna_hpo.TrialDetails.result_value_type = json_object["result_value_type"]
+                        optuna_hpo.TrialDetails.result_value = json_object["result_value"]
+                        optuna_hpo.TrialDetails.trial_result_received = 1
```
dinogun, line 70: Same for all of these.

hyperparameter_tuning/service.py
```diff
@@ -0,0 +1,124 @@
...
+                else:
+                    self._set_response(400, "-1")
+            else:
+                self._set_response(400, "-1")
+        else:
+            self._set_response(403, "-1")
+
+    def do_GET(self):
+        # TODO: perform better regex search
+        if re.search('/experiment_trials', self.path):
```
dinogun, line 83: I'd suggest that you move all of the GET and POST code into separate functions and have only the error handling code here.

dinogun (changes requested):

hyperparameter_tuning/bayes_optuna/optuna_hpo.py
```diff
@@ -78,15 +98,15 @@ def __call__(self, trial):
         return actual_sla_value
 
 
-def recommend(application_name, direction, hpo_algo_impl, id, objective_function, tunables, value_type):
+def recommend(application_name, direction, hpo_algo_impl, id_, objective_function, tunables, value_type):
```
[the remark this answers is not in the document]
agarwalrounak: `id` is the name of a [built-in function](https://docs.python.org/3/library/functions.html#id) in Python. 

In PEP 8 - Style Guide for Python Code, the following guidance appears in the section [Descriptive: Naming Styles](https://www.python.org/dev/peps/pep-0008/#descriptive-naming-styles):

> single_trailing_underscore_: used by convention to avoid conflicts with Python keyword

Thus, `id` is changed to `id_` here.

dinogun (approved):
I'll approve this for now. It needs to handle constants in the code better (eg. the various HTTP return codes), but we can do that in a subsequent PR.

PR #250, opened 2021-05-10
Add hpo_algo_impl and step and relevant yaml changes.

dinogun:


src/main/java/com/autotune/dependencyAnalyzer/application/Tunable.java
```diff
@@ -68,12 +68,22 @@ public Tunable(String name,
...
+
+		/*
+		 * Bounds cannot be negative.
+		 * upperBound has to be greater than lowerBound.
+		 * step has to be lesser than or equal to the difference between the two bounds.
+		 */
+		if (upperBound < 0 ||
+			lowerBound < 0 ||
+			lowerBound >= upperBound ||
+			step > (upperBound - lowerBound)
```
chandrams, line 81: In this case bounds might be valid, but step is invalid - should we throw an InvalidBoundsException or an InvalidStepException ?

dinogun: Step and Bounds go together and there is really only one exception. It could be for example, the upperbound was miss typed to be smaller than intended.

PR #266, opened 2021-06-22
Remove quotes and M from the test yamls ans scripts

shruacha1234:
Signed-off-by: Shruthi <Shruthi.Acharya@ibm.com>

### DIFF
--- a/tests/autotune_test_yamls/manifests/da/autotune_config_yaml/layer_presence_query/blank-layer-presence-query.yaml
+++ b/tests/autotune_test_yamls/manifests/da/autotune_config_yaml/layer_presence_query/blank-layer-presence-query.yaml
@@ -15,8 +15,8 @@ tunables:
 - name: MaxInlineLevel
   description: 'maximum number of nested calls that are inlined. Default is 9'
   value_type: integer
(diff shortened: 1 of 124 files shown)


tests/autotune_test_yamls/manifests/da/autotune_config_yaml/layer_presence_query/blank-layer-presence-query.yaml
```diff
@@ -15,8 +15,8 @@ tunables:
 - name: MaxInlineLevel
   description: 'maximum number of nested calls that are inlined. Default is 9'
   value_type: integer
-  upper_bound: "50"
-  lower_bound: "9"
+  upper_bound: 50
+  lower_bound: 9
```
chandrams, line 19: step is missing here and in other yamls that use hotspot config layer. As discussed, I think it would be good to dynamically create these yamls in results directory before the test run using the yamls in autotune/manifests. For now, I 'm ok to have these yamls updated and later include a PR.

chandrams (approved):
Changes look fine, as mentioned we can have another PR that creates yamls dynamically.

PR #280, opened 2021-08-05
Adds EM API structures for Create Experiment, Get Experiments and Status

bharathappali:
This PR is related to #279 and needs to be merged after #279 

Signed-off-by: bharathappali <bharath.appali@gmail.com>

PR #282, opened 2021-08-30
Added EM Process Queue handler and Stage Transition 

bharathappali:
This PR is based on #280 can be reviewed once #280 is merged

### DIFF
--- a/src/main/java/com/autotune/experimentManager/data/EMStageProcessQueue.java
+++ b/src/main/java/com/autotune/experimentManager/data/EMStageProcessQueue.java
@@ -1,0 +1,88 @@
+/*******************************************************************************
+ * Copyright (c) 2022, 2022 Red Hat, IBM Corporation and others.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


bharathappali:
Description: 

As #280 adds the endpoints to receive the config we need a queue structure to push the transitions to be done so that the Stage Processor can read from the queue and perform the transition. This PR adds the Process Queues and Stage Transition classes which will be pushed into the queue for the Stage Processor to read

bharathappali:
@dinogun This PR is ready to be reviewed as #280 has been merged

src/main/java/com/autotune/experimentManager/data/EMStageProcessQueue.java
```diff
@@ -0,0 +1,76 @@
+/*******************************************************************************
+ * Copyright (c) 2021, 2021 Red Hat, IBM Corporation and others.
```
dinogun, line 2: Should be 2022, 2022. Same for other files

src/main/java/com/autotune/experimentManager/data/EMStageProcessQueue.java
```diff
@@ -0,0 +1,76 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *******************************************************************************/
+
+
+package com.autotune.experimentManager.data;
+
+import java.util.concurrent.ConcurrentLinkedQueue;
+
+/*
```
dinogun, line 22: Comment should be in javadoc format. Same for other files too

dinogun (changes requested):

src/main/java/com/autotune/experimentManager/data/EMStageProcessQueue.java
bharathappali: Updated comments with java doc format

dinogun (approved):
LGTM

PR #291, opened 2021-09-30
Move to java 17 (!) and add the HPO start script.

dinogun:
- Use the eclipse temurin based maven java 17 image for builds
- Update ubi-minimal to 8.4
- Add script to start both python and Java servers.

Signed-off-by: Dinakar Guniguntala dgunigun@redhat.com

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,7 +16,7 @@
 ##########################################################
 #            Build Docker Image
 ##########################################################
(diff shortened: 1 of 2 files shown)


chandrams (approved):
I have tested it on github hosted runners, it works fine. We can go ahead and merge these changes and add any additional messages if required in another PR.

dinogun:
Will delete this PR as this is now no longer required per #320

PR #295, opened 2021-10-04
Edit API's to latest changes

bharathappali:


src/main/java/com/autotune/experimentManager/utils/EMConstants.java
```diff
@@ -1,13 +1,6 @@
 package com.autotune.experimentManager.utils;
 
 public class EMConstants {
-    public static class APIPaths {
-        public static final int PORT = 8080;
-        public static final String ROOT = "/";
-        public static final String CREATE_EXPERIMENT = ROOT + "createExperimentTrial";
-        public static final String LIST_EXPERIMENT_TRIAL = ROOT + "listExperimentTrial";
-        public static final String LIST_TRIAL_STATUS = ROOT + "listTrialStatus";
-    }
```
dinogun, line 10: Looks like you need to fixup `EMMain.Java` as well to point to the `ServerContext.Java` after this change.

bharathappali: I feel `EMMain.java` is not required at all with the EM integrated to `Autotune.java` . Earlier when I was running EM sample tests I have created EMMain which has the main class which I used tor run. Thanks for pointing out I have now removed EMMain from the code base.

dinogun (approved):
lgtm

PR #296, opened 2021-10-05
Included a PR check to build and deploy autotune integration image (RM-EM)

chandrams:
Signed-off-by: Chandrakala Subramanyam <csubrama@redhat.com>

### DIFF
--- a/test-on-pr.yaml
+++ b/test-on-pr.yaml
@@ -1,0 +1,90 @@
+name: Test on Pull Request 
+
+# Controls when the action will run. 
+on:
+  pull_request:
+    paths-ignore:
+      - 'docs/**'
+      - 'design/**'
+      - 'examples/**'
+      - 'extensions/**'
+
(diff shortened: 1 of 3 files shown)


test-on-pr.yaml
```diff
@@ -0,0 +1,90 @@
...
+      - name: Start minikube
+        run: |
+          kubectl cluster-info
+          kubectl get pods -n kube-system
+
+      - name: Build autotune
+        run: |
+          echo Build autotune
+          cd autotune
+          ./build.sh -t autotune:e2e
```
dinogun, line 50: I would prefer if we have a better name than `e2e`, maybe `integration-test` or `int-test` for short.

chandrams: I have changed it to int-test

dinogun (approved):
Thanks @chandrams. This is really good to have test automation now on the demo branch !

PR #315, opened 2021-10-29
Send summary in ListTrialStatus API

bharathappali:
This PR adds sample metric results to each of the metrics provided in the input json (metrics section in deployments).

PS: It adds same metrics results to every metric mentioned. Let me know if we need to populate different metrics for different metric names, Will implement a metric name check and add the results accordingly. Currently i was having only response time data so added the same to every metric mentioned in the metrics list.

Signed-off-by: bharathappali <bharath.appali@gmail.com>


bharathappali:
Update the sample metrics to match the appropriate queries.

dinogun:
There are several issues with this PR
1. None of the commits here are GPG signed.
2. As I have pointed out previously, please use your RH id for `signed -off-by`. Note that the GPG key should for the same email-id and it should match
3. This PR was supposed to be for summary metrics alone, not sure why it should pull in changes in 46 files.

Please fix and create a new PR with only the relevant changes.

PR #299, opened 2021-10-07
Add JSON schema validation for RM POST request for trial generate

agarwalrounak:
Closes #230

PR #320, opened 2021-11-15
Create separate containers for Java and Python

dinogun:
Update build and deploy scripts to support the same.
We will still have a single pod that will house both containers to enable the service connections to work the same.


Dockerfile.optuna
```diff
@@ -0,0 +1,61 @@
+#
+# Copyright (c) 2020, 2021 Red Hat, IBM Corporation and others.
```
chandrams, line 2: Should 2020 be 2021?

deploy.sh
```diff
@@ -121,8 +126,8 @@ do
 	n)
 		autotune_ns="${OPTARG}"
 		;;
-	d)
-		AUTOTUNE_CONFIGMAPS="${OPTARG}"
+	o)
+		OPTUNA_DOCKER_IMAGE="${OPTARG}"
```
chandrams, line 130: Extra tab / space present

chandrams:
@dinogun - Please fix the PR check

chandrams:
Tried it manually, build worked but deploying autotune failed with CrashLoopBackOff on my minikube setup

dinogun:
It looks like #299 was missing a dependency and the PR check did not catch it as the error got lost with the java messages. Infact this PR helped to point it out! Anyway it is now fixed by adding the `jsonschema` package in the pip install.

dinogun:
Hmmm on closer look, #299 did add that dependency, but it got missed in my PR as it was not synced, my bad.

chandrams (approved):
LGTM

PR #327, opened 2021-11-19
Updated test scripts with updated deploy script options

chandrams:
Updated test scripts with updated deploy script options for optuna and autotune images


dinogun (approved):
LGTM

PR #331, opened 2021-11-22
Updated HPO tests to start the HPO service when specified

chandrams:
Updated HPO tests to start the HPO service when specified


dinogun (approved):
LGTM

PR #356, opened 2021-12-08
Add demo metrics

bharathappali:
The current PR adds the demo metrics for the summary on `listTrialStatus` api.

Currently parking this PR as work-in-progress as I need to add the tested code for `completeStatus` param on `listTrialStatus` API which gives the warm up and measurement cycle data as well


bharathappali:
Sorry closing it as i need to raise against integration

PR #365, opened 2021-12-20
Updated Analyzer autotune id tests and other layer config tests

chandrams:
Updated Analyzer autotune id tests and other layer config tests

### DIFF
--- a/tests/scripts/da/autotune_id_tests.sh
+++ b/tests/scripts/da/autotune_id_tests.sh
@@ -64,7 +64,7 @@ function autotune_id_tests() {
 	fi
 	
 	# Cleanup autotune
-	autotune_cleanup  | tee -a ${LOG}
+#	autotune_cleanup  | tee -a ${LOG}
 	
 	end_time=$(get_date)
 	elapsed_time=$(time_diff "${start_time}" "${end_time}")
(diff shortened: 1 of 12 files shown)


tests/scripts/da/autotune_id_tests.sh
```diff
@@ -64,7 +64,7 @@ function autotune_id_tests() {
 	fi
 	
 	# Cleanup autotune
-	autotune_cleanup  | tee -a ${LOG}
+#	autotune_cleanup  | tee -a ${LOG}
```
dinogun, line 67: Is this no longer needed, if so, please remove it.

tests/scripts/da/autotune_layer_config_id_tests.sh
```diff
@@ -327,6 +302,14 @@ function validate_autotune_tunables_api() {
 			if [ "${config_id_test_name}" == "${autotune_layer_config_id_tests[3]}" ]; then
 				kubectl delete -f ${new_config_yaml} -n monitoring >> ${LOG}
 			fi
+
+			echo "Deleting autotune layer config yaml ${test_layer_config}..." | tee -a ${LOG}
+			kubectl delete -f ${config_yaml} -n monitoring | tee -a ${LOG}
```
dinogun, line 307: Ideally the `monitoring` ns needs to be set as a config variable. It should be set based on the cluster type (Eg. minikube, openshift etc)

dinogun (changes requested):

tests/scripts/da/autotune_id_tests.sh
chandrams: It is needed, had commented it while testing have removed the comment.

tests/scripts/da/autotune_layer_config_id_tests.sh
chandrams: Thanks, updated it.

dinogun (approved):
LGTM

PR #373, opened 2022-01-05
Support query array for layer presence query

dinogun:
1. Add a new LayerPresenceQuery class
2. Changes to Config CRD to accommodate queries array
3. Update all AutotuneConfig yamls to conform to new CRD
4. Code changes to loop through the queries and match the right k8s pods

### DIFF
--- a/manifests/autotune-configs/hotspot-micrometer-config.yaml
+++ b/manifests/autotune-configs/hotspot-micrometer-config.yaml
@@ -7,7 +7,7 @@ layer_level: 1
 details: hotspot tunables
 layer_presence:
   queries:
-  - datasource: prometheus
+  - datasource: 'prometheus'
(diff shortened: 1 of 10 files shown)


manifests/autotune-configs/hotspot-micrometer-config.yaml
```diff
@@ -5,22 +5,20 @@ metadata:
...
 details: hotspot tunables
-layerPresence:
-  query:
-    datasource:
-    - name: prometheus
-      query: 'jvm_memory_used_bytes{area="heap",}'
-      key: pod
+layer_presence:
+  queries:
+  - datasource: prometheus
```
chandrams, line 10: Prometheus does not have single quotes here. Do we need id="Tenured Gen", I 'm updating the tests in the tests it is present. Should I remove it?

chandrams (approved):
LGTM

PR #385, opened 2022-01-24
Updated autotune config yamls based on changes to autotune config CRD

chandrams:
Updated autotune config yamls based on changes to autotune config CRD for layer_presence and datasource, tunable queries (#373)


dinogun (approved):
LGTM

PR #389, opened 2022-01-27
Updated query url based on changes in PR 362

chandrams:
Updated query url based on changes in PR 362


chandrams:
Cancelling this one, will raise it against master

PR #391, opened 2022-01-28
Support query array for layer presence query

dinogun:
This PR moves the code from integration to master

1. Add a new LayerPresenceQuery class
2. Changes to Config CRD to accommodate queries array
3. Update all AutotuneConfig yamls to conform to new CRD
4. Code changes to loop through the queries and match the right k8s pods


chandrams (approved):
LGTM

PR #400, opened 2022-02-15
Adds the classes required for extracting the results from output json

bharathappali:
The class `EMMetricResult` can be used to arrange the results in the java object structure, The user of this class needs to provide the JSON object which has both or either of the keys `general_info` and `percentile_info` at the higher level of the JSONObject

example JSON to be passed:
```
{
  "general_info": {
    "mean": 95,
    "mode": 87,
    "spike": 156,
    "error": 0,
    "score": 12
  },
  "percentile_info":  {
    "50": 800,
    "95": 800,
    "97": 800,
    "99.99": 800,
    "99.999": 800
  }
}
```

### DIFF
--- a/src/main/java/com/autotune/shared/data/interfaces/ConvertToJSON.java
+++ b/src/main/java/com/autotune/shared/data/interfaces/ConvertToJSON.java
@@ -1,0 +1,7 @@
+package com.autotune.shared.data.interfaces;
+
+import org.json.JSONObject;
+
+public interface ConvertToJSON {
+    public JSONObject toJSON();
+}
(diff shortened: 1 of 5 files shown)


bharathappali:
I have placed the constants in  `AutotuneConstants` as I felt it's used by both `EM` and `Analyser`, Please lemme know if there are any comments on this change. Once we finalize will remove the constants at the other place (`AutotuneConstants` or `EMConstants`)

bharathappali:
Need to add `min`and `max` will make the chnages and push to the branch

bharathappali:
Added `min` and `max` the code is now ready for review

dinogun (approved):
Merging for MVP demo

PR #409, opened 2022-02-22
Add settings option in operator crd and corresponding classes

khansaad:
Signed-off-by: Saad Khan <saakhan@redhat.com>


dinogun:
This is no longer being pursued

PR #413, opened 2022-02-23
Undoing the rolling update to get back to the old config

bharathappali:
This change will revert back to the old config of the deployment


bharathappali:
Using deep copy of the deployment object to save in ETD and use back in clean up

bharathappali:
will fix #413

bharathappali:
Changed the field `currentDeployment` to `defaultDeployment` along with respective `getter` and `setter` functions, to be more clear about it's usage and avoid confusion about the objective of the particular field.

dinogun (approved):
LGTM

PR #416, opened 2022-02-28
Send back the Experiment Id received from input JSON in the return JSON

bharathappali:

### DIFF
--- a/src/main/java/com/autotune/experimentManager/transitions/util/TransistionHelper.java
+++ b/src/main/java/com/autotune/experimentManager/transitions/util/TransistionHelper.java
@@ -42,7 +42,7 @@ public class TransistionHelper {
         public static JSONObject getMetricsJson(String runId) {
             // Need to get data from ETD and process it
             ExperimentTrialData trialData = (ExperimentTrialData) EMMapper.getInstance().getMap().get(runId);
-            JSONObject retJsonObj = new JSONObject("{\"experiment_name\":\"galaxies-autotune-min-http-response-time\"," +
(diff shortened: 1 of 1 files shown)


src/main/java/com/autotune/experimentManager/transitions/util/TransistionHelper.java
```diff
@@ -42,7 +42,7 @@ public static JSONArray getContainerConfig(String containerName, JSONArray conta
         public static JSONObject getMetricsJson(String runId) {
             // Need to get data from ETD and process it
             ExperimentTrialData trialData = (ExperimentTrialData) EMMapper.getInstance().getMap().get(runId);
-            JSONObject retJsonObj = new JSONObject("{\"experiment_name\":\"galaxies-autotune-min-http-response-time\"," +
+            JSONObject retJsonObj = new JSONObject("{\"experiment_name\":\""+ trialData.getConfig().getEmConfigObject().getMetadata().getApplicationName() +"\"," +
                     "\"deployments\":[{\"deployment_name\":\"galaxies-sample\"," +
```
dinogun, line 46: The deployment name also needs to be from the input.json. The deployment_name + experiment_name combination is what is used to determine the thread that needs to be woken up.

bharathappali: Sure will update it.

bharathappali: @dinogun Updated the deployment name to what's available in input JSON. Please lemme know if there are any other changes required.

dinogun (approved):
LGTM

PR #417, opened 2022-03-01
Adds the is Cycle feature to stage

bharathappali:
This change makes the stages to re-visit the same stage till
the breaking condition is met. As EM needs to record the metrics
for each cycle interval we need to run the same transistion multiple
times till the cycle needs to be stop based on a dynamic condition.

Usage of intermidiate stages is to keep track of a main stage and
intermidiate stages act as sub cycle stages, but here the requirement is
a bit different. Now we don't need multiple stages to be processed but
same stage to be scheduled multiple times so now we extend EM to support
cyclcing stages as well. So now we introduce the isCycle feature in EM
Stages.

PS: Lot of work needs to be done before we finalise the design of
cycling stages


src/main/java/com/autotune/experimentManager/core/EMIterationManager.java
```diff
@@ -1,23 +1,31 @@
 package com.autotune.experimentManager.core;
```
dinogun, line 1: This file needs a legal header

src/main/java/com/autotune/experimentManager/core/EMIterationManager.java
```diff
@@ -1,23 +1,31 @@
...
 
+import com.autotune.experimentManager.data.input.EMConfigObject;
 import com.autotune.experimentManager.data.iteration.EMIterationData;
-import org.json.JSONObject;
+import com.autotune.experimentManager.exceptions.EMInvalidIterationId;
 
 import java.util.ArrayList;
 
 public class EMIterationManager {
```
dinogun, line 9: Please add a javadoc here detailing what this class is meant to do

src/main/java/com/autotune/experimentManager/core/EMIterationManager.java
```diff
@@ -1,23 +1,31 @@
...
+    private ArrayList<EMIterationData> emIterationData;
 
-    public ArrayList<EMIterationData> getEmIterationData() {
+    public ArrayList<EMIterationData> getIterationDataList() {
         return emIterationData;
     }
 
-    public EMIterationManager(int iterations) {
-        this.iterations = iterations;
+    public EMIterationManager(EMConfigObject emConfigObject) {
```
dinogun, line 18: How are we validating the emConfigObject?

src/main/java/com/autotune/experimentManager/core/EMTransitionRegistry.java
```diff
@@ -1,10 +1,12 @@
 package com.autotune.experimentManager.core;
```
dinogun, line 1: File is missing legal header

src/main/java/com/autotune/experimentManager/core/EMTransitionRegistry.java
```diff
@@ -1,10 +1,12 @@
...
 
+import com.autotune.experimentManager.data.EMMapper;
+import com.autotune.experimentManager.data.ExperimentTrialData;
 import com.autotune.experimentManager.utils.EMUtil;
 
 public class EMTransitionRegistry {
```
dinogun, line 7: Add a javadoc describing the class

src/main/java/com/autotune/experimentManager/core/EMTransitionRegistry.java
```diff
@@ -16,6 +18,22 @@ public static EMUtil.EMExpStages getNextStage(EMUtil.EMExpStages currentStage) {
...
+                return currentStage;
+            }
+        }
+        return EMUtil.EMExpStages.get(currentStage.ordinal() + 1);
+    }
+
+
     public static EMUtil.EMExpStages getPossibleContinuousStage(EMUtil.EMExpStages currentStage) {
         EMUtil.EMExpStages previousStage = currentStage;
         for (int i = currentStage.ordinal() + 1; i < EMUtil.EMExpStages.getSize(); i++) {
```
dinogun, line 39: Wondering if we are adding 1 as it starts from 0, then should it not be `<=` getSize?

bharathappali: Thanks for the review @dinogun 

A little background for the method:

The `getPossibleContinuousStage` is added to handle a future scenerio where if there are two continous independent stages which can be processed parallel ignoring if they have any intermediate `scheduled` stage. This usecase is not occured yet but EM is designed to have a handler for such scenerios.

> if we are adding 1 as it starts from 0

we are adding 1 to check from the next stages of the current stage.

> then should it not be <= getSize?

I would rather leave it with `<` getSize because we might hit an index out of range exception if `i` is equal to getSize.
The next instruction checks if `list[i]` is a scheduled transition

Note: Currently there is no usage in EM for `getPossibleContinuousStage`

PR #437, opened 2022-05-19
Fix kubectl kustomize help stdout redirect error

bharathappali:

### DIFF
--- a/scripts/minikube-helpers.sh
+++ b/scripts/minikube-helpers.sh
@@ -101,7 +101,7 @@ function check_prometheus_installation() {
 	kubectl_tool=$(which kubectl)
 	check_err "Error: Please install the kubectl tool"
 	# Check to see if kubectl supports kustomize
-	kubectl --help | grep "kustomize" > /dev/null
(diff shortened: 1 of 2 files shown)


scripts/minikube-helpers.sh
```diff
@@ -101,7 +101,7 @@ function check_prometheus_installation() {
 	kubectl_tool=$(which kubectl)
 	check_err "Error: Please install the kubectl tool"
 	# Check to see if kubectl supports kustomize
-	kubectl kustomize --help >/dev/null 2>/dev/null
+	kubectl --help | grep "kustomize" > /dev/null
```
dinogun, line 104: Bash sometimes interprets the space after '>`, so I suggest you change it to the below in both places
```
kubectl --help | grep "kustomize" >/dev/null
```

bharathappali: Thanks for the review @dinogun I have made the changes accordingly.

dinogun (approved):
LGTM

PR #444, opened 2022-06-17
Add UISM (User Interface Support Module) to autotune to support UI only related API's

bharathappali:
There are basic requirements for UI to be functional with autotune. There is a need for api's to list out `namespaces` and `Deployments` in namespace etc. This PR provides support for get Namespaces.

@dinogun Can I please have your review on this?

### DIFF
--- a/manifests/autotune-operator-role.yaml
+++ b/manifests/autotune-operator-role.yaml
@@ -19,6 +19,7 @@ rules:
   - endpoints
   - persistentvolumeclaims
   - events
+  - namespaces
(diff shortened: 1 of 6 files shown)


msvinaykumar:
It would be nice if we call node js modules inside servlet and return response.
https://github.com/kubernetes-client/javascript

Right now Im thinking we are rewriting code in each servlet for getting deployments , pods etc

dinogun:
Closing as this is stale. Re-open if needed.

PR #450, opened 2022-07-01
Add openj9 layer config

kusumachalasani:
Signed-off-by: kusumachalasani <kchalasa@redhat.com>

### DIFF
--- a/manifests/autotune-configs/openj9-actuator-config.yaml
+++ b/manifests/autotune-configs/openj9-actuator-config.yaml
@@ -24,30 +24,55 @@ layer_presence:
     key: pod
 
 tunables:
-- name: MaxInlineLevel
-  description: 'maximum number of nested calls that are inlined. Default is 9'
-  value_type: integer
-  upper_bound: '50'
-  lower_bound: '9'
-  queries:
-  - datasource: 'prometheus'
-    query: 'jvm_memory_used_bytes{area="heap", $CONTAINER_LABEL$="", $POD_LABEL$="$POD$"}'
-  step: 1
-  slo_class:
-  - response_time
-  - throughput
-
 - name: gc
   description: 'Garbage collection policy'
   value_type: categorical
   choices:
-    - gencon
-    - balanced
-    - optavgpause
-    - optthruput
-    - metronome
+    - 'gencon'
+    - 'balanced'
+    - 'optavgpause'
+    - 'optthruput'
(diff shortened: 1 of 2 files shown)


manifests/autotune-configs/openj9-actuator-config.yaml
```diff
@@ -0,0 +1,53 @@
...
+tunables:
+- name: MaxInlineLevel
+  description: 'maximum number of nested calls that are inlined. Default is 9'
+  value_type: integer
+  upper_bound: '50'
+  lower_bound: '9'
+  queries:
+  - datasource: 'prometheus'
+    query: 'jvm_memory_used_bytes{area="heap", $CONTAINER_LABEL$="", $POD_LABEL$="$POD$"}'
+  step: 1
```
dinogun, line 35: Please move this to the GC policy as that makes more sense. Can you also make the change in hotspot as well

manifests/autotune-configs/openj9-actuator-config.yaml
```diff
@@ -0,0 +1,53 @@
...
+
+- name: gc
+  description: 'Garbage collection policy'
+  value_type: categorical
+  choices:
+    - gencon
+    - balanced
+    - optavgpause
+    - optthruput
+    - metronome
```
dinogun, line 48: Does metronome work for all JDK's ? Or is it only for special builds?

manifests/autotune-configs/openj9-actuator-config.yaml
kusumachalasani: Done.

dinogun (commented):
@kusumachalasani Have you tested both openj9 and hotspot images to make sure
1. The right layers are being matched with the right images
2. It is not detecting both layers at once

kusumachalasani:
> @kusumachalasani Have you tested both openj9 and hotspot images to make sure
> 
> 1. The right layers are being matched with the right images
> 2. It is not detecting both layers at once

Yes, I did test by running TFB benchmark with hotspot and petclinic with openj9. Both of them detected the layers correctly.

manifests/autotune-configs/openj9-actuator-config.yaml
kusumachalasani: it is working for all builds.

dinogun (approved):
LGTM

PR #453, opened 2022-07-02
Remove quarkus.http.io-threads and virtualized tunable

kusumachalasani:
Remove quarkus.http.io-threads tunable from hotspot and virtualized tunable from openj9 to avoid deployment crash as they require code changes.


dinogun (approved):
Approving for now as these require Java code changes for it to work.

PR #469, opened 2022-08-02
Remove kube ref

msvinaykumar:
Remove Kube client from AutotuneDeployment.java so that modules are loosely coupled with java Kube client.


dinogun:
Please fix the conflicts

PR #471, opened 2022-08-02
Get kube client design

msvinaykumar:
After analysis decide if getting Kube client have to be Singleton.


dinogun:
Please rebase to the latest branch

PR #482, opened 2022-08-04
Dummy loop for A/B testing.

msvinaykumar:
Signed-off-by: msvinaykumar <vinakuma@redhat.com>

PR #484, opened 2022-08-08
Dummy metric generator for autotune api's

bharathappali:
NOTE: This PR is build on top of #482 so it needs to be merged before this gets merged

This PR has the respective caller function to return the dummy api.

Changes made:
- the experiment trial is now stored in EM Mapper with a new map which holds the experiment name and trial mapping of a particular experiment trial object (needed to construct a return json which has the trial details)
- a dummy json function which generates the return json based on the experiment name, trail and verboseness

### DIFF
--- a/src/main/java/com/autotune/experimentManager/core/ExperimentTrialHandler.java
+++ b/src/main/java/com/autotune/experimentManager/core/ExperimentTrialHandler.java
@@ -19,6 +19,7 @@ import com.autotune.common.experiments.ExperimentTrial;
 import com.autotune.common.experiments.PodContainer;
 import com.autotune.common.target.kubernetes.service.KubernetesServices;
 import com.autotune.common.target.kubernetes.service.impl.KubernetesServicesImpl;
+import com.autotune.experimentManager.data.EMMapper;
 import com.autotune.utils.HttpUtils;
 import com.google.gson.Gson;
 import org.json.JSONArray;
@@ -30,6 +31,8 @@ import java.net.MalformedURLException;
 import java.net.URL;
 import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.List;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.stream.IntStream;
 
 import static com.autotune.analyzer.loop.HPOInterface.postTrialResultToHPO;
@@ -44,6 +47,19 @@ public class ExperimentTrialHandler {
 
     public ExperimentTrialHandler(ExperimentTrial experimentTrial) {
         this.experimentTrial = experimentTrial;
+        String experimentName = experimentTrial.getExperimentName();
+        String trialNum = String.valueOf(experimentTrial.getTrialInfo().getTrialNum());
+        ConcurrentHashMap<String, HashMap<String, ExperimentTrial>> expTrialMap = EMMapper.getInstance().getExpTrialMap();
+        if (expTrialMap.containsKey(experimentName)) {
+            HashMap<String, ExperimentTrial> expMap = expTrialMap.get(experimentName);
+            if (!expMap.containsKey(trialNum)) {
+                expMap.put(trialNum, experimentTrial);
+            }
+        } else {
+            HashMap<String, ExperimentTrial> trialMap = new HashMap<String, ExperimentTrial>();
+            trialMap.put(trialNum, experimentTrial);
+            expTrialMap.put(experimentName, trialMap);
+        }
     }
 
     public static JSONObject getDummyMetricJson(ExperimentTrial experimentTrial) {
--- a/src/main/java/com/autotune/experimentManager/data/EMMapper.java
+++ b/src/main/java/com/autotune/experimentManager/data/EMMapper.java
@@ -1,6 +1,10 @@
 package com.autotune.experimentManager.data;
 
+import com.autotune.common.experiments.ExperimentTrial;
+
+import java.util.HashMap;
 import java.util.LinkedList;
+import java.util.List;
 import java.util.concurrent.ConcurrentHashMap;
 
 public class EMMapper {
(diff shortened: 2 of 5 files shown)


src/main/java/com/autotune/experimentManager/core/ExperimentTrialHandler.java
```diff
@@ -44,6 +47,19 @@ public class ExperimentTrialHandler {
 
     public ExperimentTrialHandler(ExperimentTrial experimentTrial) {
         this.experimentTrial = experimentTrial;
+        String experimentName = experimentTrial.getExperimentName();
+        String trialNum = String.valueOf(experimentTrial.getTrialInfo().getTrialNum());
+        ConcurrentHashMap<String, HashMap<String, ExperimentTrial>> expTrailMap = EMMapper.getInstance().getExpTrialMap();
```
msvinaykumar, line 52: Can you store like this
 ```
         { 
                  "expName1" :  ExperimentTrialObject1 ,
                  "expName2" :  ExperimentTrialObject2 ,
          }    
```
and inside ExperimentTrail object navigate to find if trail number exist. if exist return 400 else 200  . Please move this logic to some class and call this class in CreateExperimentTrial servlet. 

      
        ` this.experimentTrial.getTrialDetails().forEach((tracker, trialDetails) -> {
            trialDetails.getPodContainers().forEach((imageName, podContainer) -> {
                podContainer.getTrialConfigs().forEach((trialNumber, containerConfigData`

File: src/main/java/com/autotune/experimentManager/data/EMMapper.java [hunk 2]
```diff
@@ -1,16 +1,23 @@
...
 import java.util.LinkedList;
+import java.util.List;
 import java.util.concurrent.ConcurrentHashMap;
 
 public class EMMapper {
     private static volatile EMMapper emMapper = null;
     private volatile ConcurrentHashMap<String, ExperimentTrialData> emMap = null;
     private volatile ConcurrentHashMap<String, LinkedList<String>> emDeploymentRunIdMap = null;
 
+    private volatile ConcurrentHashMap<String, HashMap<String, ExperimentTrial>> emExpTrialMap = null;
```
msvinaykumar, line 15: Change it to

    private volatile ConcurrentHashMap<String,  ExperimentTrial> emExpTrialMap = null;

src/main/java/com/autotune/experimentManager/data/EMMapper.java
```diff
@@ -1,16 +1,23 @@
...
+
     private EMMapper() {
         emMap = new ConcurrentHashMap<String, ExperimentTrialData>();
         emDeploymentRunIdMap = new ConcurrentHashMap<String, LinkedList<String>>();
+        emExpTrialMap = new ConcurrentHashMap<String, HashMap<String, ExperimentTrial>>();
```
msvinaykumar, line 20: Change it to

    private volatile ConcurrentHashMap<String,  ExperimentTrial> emExpTrialMap = null;

src/main/java/com/autotune/experimentManager/core/ExperimentTrialHandler.java
```diff
@@ -44,6 +47,19 @@ public class ExperimentTrialHandler {
...
+                expMap.put(trialNum, experimentTrial);
+            }
+        } else {
+            HashMap<String, ExperimentTrial> trialMap = new HashMap<String, ExperimentTrial>();
+            trialMap.put(trialNum, experimentTrial);
+            expTrailMap.put(experimentName, trialMap);
+        }
     }
 
     public static JSONObject getDummyMetricJson(ExperimentTrial experimentTrial) {
```
msvinaykumar, line 65: Can you populate dummy values by calling empty abstract functions for each warm up and measurement cycles .
so tomorrow We just have to modify empty abstract functions to get real metric values. 
Look for comment here.
`        this.experimentTrial.getTrialDetails().forEach((tracker, trialDetails) -> {
            trialDetails.getPodContainers().forEach((imageName, podContainer) -> {
                podContainer.getTrialConfigs().forEach((trialNumber, containerConfigData) -> {
                    DeploymentHandler deploymentHandler = new DeploymentHandler(
                            trialDetails.getDeploymentNameSpace(),
                            trialDetails.getDeploymentName(),
                            containerConfigData);
                    IntStream.rangeClosed(1, numberOFIterations).forEach(
                            i -> {
                                deploymentHandler.initiateDeploy();
                                //check if load applied to deployment
                                //collect warmup and measurement cycles metrics
                            }
                    );
                });
            });
        });`

src/main/java/com/autotune/experimentManager/services/ListTrialStatus.java
```diff
@@ -38,56 +39,20 @@
...
-        ArrayList<String> runIdList = new ArrayList<String>();
-        boolean validRunId = true;
-        String runIdParam = req.getParameter(EMConstants.InputJsonKeys.ListTrialStatusKeys.RUN_ID);
-        if (null != runIdParam) {
-            runIdList.add(runIdParam);
-        } else {
-            EMMapper.getInstance().getMap().forEach((key, value) -> {
-                runIdList.add((String) key);
-            });
+        String experiment_name = null;
```
msvinaykumar, line 42: Can you handle exception Record not found 404 and others if any like 400

src/main/java/com/autotune/experimentManager/services/util/EMAPIHandler.java
```diff
@@ -81,4 +87,162 @@ private static void pushTransitionToQueue(String runId) {
...
+                resJson.put("Status", "COMPLETED");
+                returnJson.put(trialNum, resJson);
+                return returnJson;
+            }
+        }
+
+        return returnJson;
+    }
+
+    private static JSONObject getDummyTrialJSON(ExperimentTrial experimentTrial, boolean verbose) {
```
msvinaykumar, line 128: Seems duplicate function appeared twice .

msvinaykumar:
Please add java documentation .

src/main/java/com/autotune/experimentManager/core/ExperimentTrialHandler.java
bharathappali: I'm expecting we will lose the track of the older trial's if we maintain it in a map of exp name and trial object, From my understanding every experiment has multiple trials identified by trial num. If i store exp 1 and trial 0 in the map and when trial 1 gets created the map is updated with trial 1 and details of the trial 0 will be lost (unless we store it in a DB) so for now I would suggest we maintain it in a exp name and list of trials or map of trials (which I have choose to identify a trial object based on exp name and trial number).

src/main/java/com/autotune/experimentManager/data/EMMapper.java
bharathappali: I'm expecting that changing from a data structure mapping (string, hashmap) to data point mapping (string, object) makes autotune to lose the ability of storing the older trial information. Atleast till a point we have a DB to record older trial information we need to maintain it locally via global maps (which the current implementation does).

bharathappali:
I feel the data structure for metric mapping already exists in ETD (Experiment Trial Data) It would be hassle free if we place a pointer for `ExperimentTrial` in ETD by replacing `EMConfig` as both of them are identical (holds the input JSON)

dinogun (approved):
Merging this into the demo branch for simulating the dummy metrics

PR #486, opened 2022-08-17
Change in input JSON template to cover main usecases.

msvinaykumar:
JSON template change to cover following use cases.

Use Cases

EM currently is designed to handle the following **use cases**

**- A/B testing**

Who: Dev
Where: Part of the CI/CD pipeline
Trigger: Source code changes / commit
In this usecase, the user provides a deployment name, two docker images and metrics to be monitored during the course of the experiment. The goal is to deploy the docker images in two separate trials respectively, one after the other. It will monitor the given metrics during both the trials and report the results at the end of each trial.
The results from each of the trial can then be used to compare the two trials and determine if there has been a variation in the performance objective and the quantum of the same.

**- Run experiment trials in a Dev/QA environment and generate a recommendation**

Who: Dev
Where: Part of the Dev/QA environment
Trigger: A/B testing reveals a huge variation which necessitates a fresh Autotune run to generate new config that meets the performance objectives. It could also be for a new project.
In this usecase, the user provides a deployment name and a performance objective. The performance objective consists of metrics to be monitored over the course of the experiment. The goal is to find the config supplied by HPO that best optimizes the user provided objective. The config can consist of both resource requests and limits and env parameters. In this case, from a EM point of view, this will usually translate to many trials, usually a 100, that each have a different config that needs to be deployed (using a rolling update) and monitor the metrics provided. The results will be returned for each such trial.

**- Monitor a deployment over long term and provide a recommendation.**

Who: SRE
Where: Production
Trigger: Better understand production load to reduce costs
In this usecase, the SRE will provide a deployment that needs to be monitored over a long period of time. The intention is to better understand the variance of the incoming load conditions and provide recommendations on the container and heap sizing in an effort to reduce costs. From the EM point of view, this will translate into monitoring an existing deployment over long periods on the user specified metrics and updated usage results on a periodic basis.


examples/EMTemplates/EM_ABTesting_Template.json
```diff
@@ -0,0 +1,86 @@
+[
+  {
+    "experiment_name": "Compare performance objective of A/B software releases",
+    "resource": {
```
msvinaykumar, line 4: We might have to add "containername" here

examples/EMTemplates/EM_ABTesting_Template.json
```diff
@@ -0,0 +1,86 @@
...
+    "settings": {
+      "trial_settings": {
+        "measurement_cycles": "3",
+        "warmup_duration": "1min",
+        "warmup_cycles": "3",
+        "measurement_duration": "1min",
+        "iterations": "3"
+      },
+      "deployment_settings": {
+        "deployment_tracking": {
```
msvinaykumar, line 23: Shall we keep this or remove for now .

examples/EMTemplates/EM_GeneralPerfExperiments_Template.json
```diff
@@ -0,0 +1,206 @@
+[
+  {
+    "experiment_name": "quarkus-resteasy-autotune-min-http-response-time-db",
+    "resource": {
```
msvinaykumar, line 4: can we add 
"containerName" and 
"imageName" here as it remain constant across all trials .

examples/EMTemplates/EM_GeneralPerfExperiments_Template.json
```diff
@@ -0,0 +1,206 @@
...
+          "query": "rate(http_server_requests_seconds_sum{method=\"GET\",outcome=\"SUCCESS\",status=\"200\",uri=\"/db\",}[1m])",
+          "name": "request_sum"
+        },
+        "request_count": {
+          "datasource": "prometheus1",
+          "query": "rate(http_server_requests_seconds_count{method=\"GET\",outcome=\"SUCCESS\",status=\"200\",uri=\"/db\",}[1m])",
+          "name": "request_count"
+        }
+      },
+      "container_metrics": {
```
bharathappali, line 46: Earlier version of EM input JSON has structured the container metrics in a array format as there might be multiple containers per pod. With the current approach that use case might be missed or we need to make EM understand the new input JSON way of detecting multiple containers. Something like passing container name as one of the attribute.

msvinaykumar: done

examples/em-jsontemplates/ABTestingTemplate.json
```diff
@@ -0,0 +1,83 @@
...
+      },
+      "request_count": {
+        "datasource": "prometheus1",
+        "query": "rate(http_server_requests_seconds_count{method=\"GET\",outcome=\"SUCCESS\",status=\"200\",uri=\"/db\",}[1m])",
+        "name": "request_count"
+      }
+    },
+    "container_metrics": {
+      "container_name": {
+        "MaxInlineLevel": {
```
dinogun, line 44: `jvm_memory_used_bytes` is no longer associated with `MaxInineLevel` but with `GC`

dinogun (commented):
As discussed, pls add the response URL section back into the input.json

dinogun (approved):
LGTM

PR #493, opened 2022-08-25
Proceed to check "isEmpty" only if the string is "not null"

bharathappali:
We should be setting the authorization header if the `authHeaderString` is not null and not empty. The earlier check used an `or case` where the program proceeds to check if it's not empty even though the string is `null` as the OR gate checks both the conditions to decide the outcome, Now we replace it with a logical `and case` where we proceed only if the first case (`not null`) case is true.

This fix avoids the null pointer exception raised when we call `isEmpty()` over a `null`


bharathappali:
@dinogun Just made a trivial change of placing `null` as left (first) operand as a better practice, Can I please have your review on it? Thanks in advance.

dinogun (approved):
LGTM

PR #495, opened 2022-08-30
Adds load interceptor mechanism

bharathappali:
The metrics collection should be started only if there is a load on the deployment. This PR adds a base implementation of the load detection.

Work which needs to be done on top of this PR:
- Check for pod metrics in experiment trial and decide load can be detected
- Check for minimal variation required in pod metrics (if available and if not alternate metrics) to decide if load is available

### DIFF
--- a/src/main/java/com/autotune/experimentManager/core/ExperimentTrialHandler.java
+++ b/src/main/java/com/autotune/experimentManager/core/ExperimentTrialHandler.java
@@ -22,15 +22,9 @@ import com.autotune.common.target.kubernetes.service.KubernetesServices;
 import com.autotune.common.target.kubernetes.service.impl.KubernetesServicesImpl;
 import com.autotune.experimentManager.core.interceptor.EMLoadInterceptor;
 import com.autotune.experimentManager.data.EMMapper;
-<<<<<<< HEAD
-<<<<<<< HEAD
+
 import com.autotune.experimentManager.utils.EMConstants;
-=======
-=======
-import com.autotune.experimentManager.utils.EMConstants;
->>>>>>> Adds the proceed to collection of metrics check
 import com.autotune.experimentManager.utils.EMUtil;
->>>>>>> Adds load interceptor mechanism
 import com.autotune.utils.HttpUtils;
 import com.google.gson.Gson;
 import org.json.JSONArray;
@@ -228,6 +222,7 @@ public class ExperimentTrialHandler {
                 );
             } catch (Exception e) {
                 LOGGER.error(e.toString());
+                e.printStackTrace();
             } finally {
                 if (kubernetesServices != null)
                     kubernetesServices.shutdownClient();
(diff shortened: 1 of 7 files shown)


src/main/java/com/autotune/experimentManager/core/ExperimentTrialHandler.java
```diff
@@ -153,27 +157,69 @@ public void startExperimentTrials() {
...
+                        for (int j = count; j > 0 && !deploymentHandler.isDeploymentReady(); j--) {   //Parameterize sleep duration hardcoded for 120
+                            try {
+                                LOGGER.debug("Still deployment is not ready for ExpName {} trail No {}", this.experimentTrial.getExperimentName(), tracker);
+                                Thread.sleep(EMConstants.TimeConv.DEPLOYMENT_CHECK_INTERVAL_IF_READY_MILLIS);
+                            } catch (InterruptedException e) {
+                                e.printStackTrace();
                             }
-                            if (!deploymentHandler.isDeploymentReady())
-                                LOGGER.debug("Giving up for ExpName {} trail No {} for {} attempt", this.experimentTrial.getExperimentName(), this.experimentTrial.getTrialInfo().getTrialNum(), i);
+                        }
```
dinogun, line 174: This whole code has to be moved into a function, ideally part of the target abstraction itself

src/main/java/com/autotune/experimentManager/core/ExperimentTrialHandler.java
```diff
@@ -153,27 +157,69 @@ public void startExperimentTrials() {
...
+                                if (this.experimentTrial.getExperimentSettings().getTrialSettings().isForceCollectMetrics()) {
+                                    proceedToMetricsCollection = true;
+                                }
+                            }
+                            if (!proceedToMetricsCollection) {
+                                LOGGER.debug("Proceeding to next trial and this trial cannot be completed due to lack of metrics collection. Will be marked as failed trial");
+                                // Need to implement a exiting functionality
+                            } else {
+                                //collect warmup and measurement cycles metrics
+                            }
```
dinogun, line 219: Why is this code here and not in the load abstraction? It should just be a simple call here to check whether we move forward or error out

src/main/java/com/autotune/experimentManager/core/ExperimentTrialHandler.java
bharathappali: I have add an else case just to check the readiness of deployment and proceed to load check. I agree these needs to be part of target abstraction but I feel @msvinaykumar might have a plan/implementation around that part. I would suggest for  a discussion with @msvinaykumar on the plan for this, and if nothing related to that is in current work plan will take it up. I just wanna make sure I'm not duplicating the effort or interrupting an existing plan / work. 

@dinogun , Please lemme know your views on it

dinogun (approved):
Since the deployment parts of the `ExperimentTrialHandler` are being handled in a separate PR, approving this

PR #499, opened 2022-09-12
Adds deployment ready abstraction

bharathappali:
This PR adds the deployment ready abstraction which loops the kubernetes service check and returns if we can proceed.

This PR is currently work in progress, The Experiment Trial object needs to be passed to the abstraction to get a better visibility of ongoing process check. 

Signed-off-by: bharathappali <abharath@redhat.com>

PR #502, opened 2022-09-13
Set flags for workflows

bharathappali:
Continuation of #499 

The EM trial flow should check if sub flows are needed like is deployment needed for a particular trial, is load check needed for a particular trial etc.


dinogun (approved):
LGTM

PR #504, opened 2022-09-14
Makes the constants final

bharathappali:
In some parts of autotune, the constants are not final which makes them editable / changeable in any place. So making the constants to be final. Also made the classes to be final making them not extendable (we could discuss around this and make changes accordingly but the constants should be marked as final to avoid changes to them)


bharathappali:
Fixes #505

dinogun (approved):
LGTM!

PR #507, opened 2022-09-14
Remove duplicate constants

### LINKED ISSUE #506: Duplicate constants
Many of the constants which are earlier EM related are now duplicated in two places (Autotune Constants and EM Constants) based on their usage in the project each individual constant should be placed and accessed from their appropriate location in the autotune project

bharathappali:
Built on top of #504 . Removes the duplicate constants in EM, fixes #506


dinogun (approved):
This was needed! LGTM

PR #516, opened 2022-09-29
Check if pods are ready.

msvinaykumar:


src/main/java/com/autotune/common/target/kubernetes/service/KubernetesServices.java
```diff
@@ -86,5 +86,9 @@ public interface KubernetesServices {
     //Watch endpoints. Used to trigger events based on some action to kubernetes resources.
     void watchEndpoints(CustomResourceDefinitionContext crd);
 
+    //Check if deployment is ready
     public boolean isDeploymentReady(String namespace, String deploymentName);
+
+    //Check if pods is ready
+    public boolean isPodsReady(String namespace, int podsReadyInMinutes);
```
dinogun, line 93: 1. If we have to check for pods of a specific deployment, deploymentName should be one of the parameters as well?
2. Naming doesn't seem right. Should it be `isPodReady`?

src/main/java/com/autotune/common/target/kubernetes/service/impl/KubernetesServicesImpl.java
```diff
@@ -511,6 +511,11 @@ public boolean isDeploymentReady(String namespace, String deploymentName) {
         return deploymentReady;
     }
 
+    @Override
+    public boolean isPodsReady(String namespace, int podsReadyInMinutes) {
+        return false;
+    }
```
dinogun, line 517: If this is a dummy implementation to be replaced with the proper check later, can we please add a `TODO` in the comments indicating that

src/main/java/com/autotune/common/target/kubernetes/service/KubernetesServices.java
```diff
@@ -86,5 +86,9 @@ public interface KubernetesServices {
     //Watch endpoints. Used to trigger events based on some action to kubernetes resources.
     void watchEndpoints(CustomResourceDefinitionContext crd);
 
+    //Check if deployment is ready
     public boolean isDeploymentReady(String namespace, String deploymentName);
+
+    //Check if pods are running
+    public boolean isPodsRunningStatus(String namespace,String deploymentName);
```
dinogun, line 93: This should be `arePodsRunning`. Space after ","

src/main/java/com/autotune/common/target/kubernetes/service/impl/KubernetesServicesImpl.java
```diff
@@ -50,6 +53,18 @@ public KubernetesServicesImpl() {
...
+        String namespace = "default";
+        KubernetesClient kubernetesClient = new DefaultKubernetesClient();
+        List<Pod> list = kubernetesClient.pods().inNamespace(namespace).list().getItems();
+        list.forEach((pods) -> {
+            System.out.println(pods.getMetadata().getName() + "---" + pods.getStatus().toString());
+            System.out.println(pods.getStatus().getPhase());
+        });
+
+    }
+
```
dinogun, line 67: Why is this needed?

src/main/java/com/autotune/common/utils/ExponentialBackOff.java
```diff
@@ -0,0 +1,190 @@
...
+import java.util.Random;
+import java.util.concurrent.ThreadLocalRandom;
+
+/**
+ * This builder class helps in retrying specific task using exponential backoff approach.
+ */
+public final class ExponentialBackOff {
+    private static final Logger LOGGER = LoggerFactory.getLogger(ExponentialBackOff.class);
+    // Maximum total time in Millis to complete exponential backoff approach. Default set to 10sec
+    public static long defaultMaxElapsedTimeMillis = 10 * 1000;   //1o Seconds
```
bharathappali, line 30: The Default values can be placed in `AutotuneConstants` with a suitable variable names to keep it consistent.

src/main/java/com/autotune/common/utils/ExponentialBackOff.java
```diff
@@ -0,0 +1,190 @@
...
+    //  increases the back off period for each retry attempt using a randomization function that grows exponentially.
+    private final double randomizationFactor;
+    private final double multiplier;
+    private int numberOfTriesLeft;
+    private long retryIntervalMillis;
+    private long randomizedIntervalMillis;
+    private double totalRetryIntervalMillis = 0;
+
+
+    public ExponentialBackOff(Builder builder) {
```
bharathappali, line 56: `null` check for `builder` can avoid Null pointer exception when it access the attributes of builder object. Also we can create a new builder object if null value is passed.

src/main/java/com/autotune/common/utils/ExponentialBackOff.java
```diff
@@ -0,0 +1,190 @@
...
+        this.retryIntervalMillis = (builder.retryIntervalMillis > 0) ? builder.retryIntervalMillis : defaultTimeToWait;
+        this.multiplier = (builder.multiplier > 0) ? builder.multiplier : defaultMultiplier;
+    }
+
+    public boolean shouldRetry() {
+        boolean shouldRetry = false;
+        if (Math.round(this.totalRetryIntervalMillis) < this.maxElapsedTimeMillis)
+            shouldRetry = true;
+        else if (numberOfRetries > 0) {
+            if (numberOfTriesLeft > 0) {
```
bharathappali, line 71: Can the both if conditions be merged into one? Something like `else if (numberOfRetries > 0 && numberOfTriesLeft > 0)`?

dinogun (approved):
LGTM

PR #528, opened 2022-10-13
Reverting time range of query

bharathappali:
Needed to revert the time range to progress demo in short period and to match the cycle duration


bharathappali:
It's a trivial change and can be merged after the build is successful

dinogun (approved):
LGTM

PR #533, opened 2022-12-01
Switching Ubuntu version to 20.04

chandrams:
Minikube github action failed on Ubuntu-22.04, switching to Ubuntu-20.04


chandrams:
Raised it against master by mistake, will raise it against mvp_demo

PR #541, opened 2022-12-15
Add a performance profile template to define queries

dinogun:
1. This is used for standard templates such as resource optimization.
2. Add a resource_optimization_openshift profile


dinogun:
Merging this into mvp_demo as this has already been discussed

PR #543, opened 2022-12-23
Included initial set of tests for Kruize monitoring APIs

chandrams:
Signed-off-by: Chandrakala Subramanyam <csubrama@redhat.com>


tests/scripts/remote_monitoring_tests/helpers/json_validate.py
```diff
@@ -0,0 +1,128 @@
...
+VALUE_MISSING = " cannot be empty or null!"
+
+exp_input_schema = {
+    "type": "array",
+    "items": {
+        "type": "object",
+        "properties": {
+            "name": {"type": "string"},
+            "namespace": {"type": "string"},
+            "slo": {
```
dinogun, line 33: What about `performanceProfile`? It can be either `slo` or `performanceProfile` that can be specified. Infact in the remote_monitoring case, it will always be `performanceProfile`

chandrams: I forgot to remove this file, will remove it. I 'm not using this for the tests, had added it only for the demo as the json validation will need to be done in the source code.

dinogun (approved):
LGTM

PR #549, opened 2023-01-19
Fixed issues related to image that running on Docker without CRD.

msvinaykumar:

### DIFF
--- a/src/main/java/com/autotune/analyzer/Analyzer.java
+++ b/src/main/java/com/autotune/analyzer/Analyzer.java
@@ -15,6 +15,7 @@
  *******************************************************************************/
 package com.autotune.analyzer;
 
+import com.autotune.analyzer.deployment.AutotuneDeploymentInfo;
 import com.autotune.analyzer.deployment.KruizeDeployment;
 import com.autotune.analyzer.deployment.InitializeDeployment;
 import com.autotune.analyzer.exceptions.K8sTypeNotSupportedException;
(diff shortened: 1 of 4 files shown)


src/main/java/com/autotune/analyzer/Analyzer.java
```diff
@@ -23,27 +24,33 @@
...
 public class Analyzer {
+    private static final Logger LOGGER = LoggerFactory.getLogger(AutotuneDeploymentInfo.class);
     public static void start(ServletContextHandler contextHandler) {
         try {
             InitializeDeployment.setup_deployment_info();
-
         } catch (Exception | K8sTypeNotSupportedException | MonitoringAgentNotSupportedException | MonitoringAgentNotFoundException e) {
-            e.printStackTrace();
-            // Current deployment not supported. Exit
-            System.exit(1);
```
dinogun, line 35: I think this needs to be done only for Monitoring related exceptions. For the other ones we still need the stack trace

src/main/java/com/autotune/analyzer/Analyzer.java
```diff
@@ -23,27 +24,33 @@
...
+          LOGGER.warn("Getting the deployment information failed.");
+        }
+        try {
+            Experimentator.start();
+        }catch (Exception e){
```
dinogun, line 40: Coding standards: Space before `{` and after `}`

src/main/java/com/autotune/analyzer/Analyzer.java
```diff
@@ -23,27 +24,33 @@
...
+            LOGGER.warn("Failed to start Experiment.");
```
dinogun, line 41: When do we expect this to hit? Is this a OOM situation?

src/main/java/com/autotune/analyzer/deployment/AutotuneDeploymentInfo.java
```diff
@@ -150,8 +150,7 @@ public static void setMonitoringAgent(String monitoringAgent) throws MonitoringA
         if (AutotuneSupportedTypes.MONITORING_AGENTS_SUPPORTED.contains(monitoringAgent)) {
             AutotuneDeploymentInfo.monitoringAgent = monitoringAgent;
         } else {
-            LOGGER.error("Monitoring agent {}  is not supported", monitoringAgent);
-            throw new MonitoringAgentNotSupportedException();
+            LOGGER.warn("Monitoring agent is mandatory for Local Monitoring/Experiment mode.");
```
dinogun, line 153: We still need to know what was the `monitoringAgent` specified. In case a monitoring agent that is not supported, we need to mention that

src/main/java/com/autotune/common/target/kubernetes/service/impl/KubernetesServicesImpl.java
```diff
@@ -403,7 +403,7 @@ public void addWatcher(CustomResourceDefinitionContext crd, Watcher watcher) {
         try {
             kubernetesClient.customResource(crd).watch(watcher);
         } catch (Exception e) {
-            new TargetHandlerException(e, "addWatcher failed!");
+            LOGGER.warn("Failed to add watcher and presuming kruize running for Remote monitoring use case. ");
```
dinogun, line 406: This is a misleading message. No Watcher means no CRD, this means REST API access only is enabled. You can still run experiments of all types but only through REST API. The message should be
```
LOGGER.warn("Failed to add Watcher. Only REST API access enabled. ");
```

PR #551, opened 2023-01-30
Update queries for resource_optimization performance profile

kusumachalasani:
Signed-off-by: kusumachalasani <kchalasa@redhat.com>


dinogun (approved):
These queries are being used with the current algo changes, hence approving.

PR #556, opened 2023-02-06
MD changes - aggregation_info

msvinaykumar:
General_info renamed to aggregation_info
Changed "Long Term" to "Performance", "Short Term" to "Cost" and "Medium Term" to "Balanced"

### DIFF
--- a/src/main/java/com/autotune/analyzer/utils/GsonUTCDateAdapter.java
+++ b/src/main/java/com/autotune/analyzer/utils/GsonUTCDateAdapter.java
@@ -33,8 +33,8 @@ public class GsonUTCDateAdapter implements JsonSerializer<Date>, JsonDeserialize
     private final DateFormat dateFormat;
 
     public GsonUTCDateAdapter() {
-        dateFormat = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSS'Z'", Locale.US);      //This is the format I need
-        dateFormat.setTimeZone(TimeZone.getTimeZone("UTC"));                               //This is the key line which converts the date to UTC which cannot be accessed with the default serializer
+        dateFormat = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSS'Z'", Locale.US);
+        dateFormat.setTimeZone(TimeZone.getTimeZone("UTC"));
     }
 
     @Override
--- a/src/main/java/com/autotune/analyzer/utils/GsonUTCDateAdapter.java
+++ b/src/main/java/com/autotune/analyzer/utils/GsonUTCDateAdapter.java
@@ -33,7 +33,7 @@ public class GsonUTCDateAdapter implements JsonSerializer<Date>, JsonDeserialize
     private final DateFormat dateFormat;
 
     public GsonUTCDateAdapter() {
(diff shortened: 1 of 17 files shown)


dinogun:
Please update any MD files as well as test related JSONs that need the format to be updated

src/main/java/com/autotune/analyzer/utils/GsonUTCDateAdapter.java
```diff
@@ -0,0 +1,53 @@
...
+
+/**
+ * UTC data handler for GSON
+ */
+public class GsonUTCDateAdapter implements JsonSerializer<Date>, JsonDeserializer<Date> {
+
+    private final DateFormat dateFormat;
+
+    public GsonUTCDateAdapter() {
+        dateFormat = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSS'Z'", Locale.US);      //This is the format I need
```
dinogun, line 36: Why is the locale here the `US`? Also what happens when we send in dates not in the UTC TZ

src/main/java/com/autotune/analyzer/utils/GsonUTCDateAdapter.java
```diff
@@ -0,0 +1,53 @@
...
+
+    @Override
+    public synchronized Date deserialize(JsonElement jsonElement, Type type, JsonDeserializationContext jsonDeserializationContext) {
+        try {
+            return dateFormat.parse(jsonElement.getAsString());
+        } catch (ParseException e) {
+            throw new JsonParseException(e);
+        }
+    }
+}
```
dinogun, line 53: Missing EoF

src/main/java/com/autotune/analyzer/utils/GsonUTCDateAdapter.java
msvinaykumar: @dinogun thanks ofr pointing it out. as per syntax we have to specify some locale and  I just changed it to root , The root locale is the locale whose language, country, and variant are empty ("") strings.

Input Json should follow this format if not throw error
if Z is not specifed  then API will throw error 
But it allows if we specify it as +00:00  
in similar fashion input request are allowed to specify UTC as

"−05:00" for New York on standard time ([UTC−05:00](https://en.wikipedia.org/wiki/UTC%E2%88%9205:00))
    "−04:00" for New York on daylight saving time ([UTC−04:00](https://en.wikipedia.org/wiki/UTC%E2%88%9204:00))
    "+00:00" (but not "−00:00") for London on standard time ([UTC±00:00](https://en.wikipedia.org/wiki/UTC%C2%B100:00))
    "+02:00" for Cairo ([UTC+02:00](https://en.wikipedia.org/wiki/UTC%2B02:00))
    "+05:30" for Mumbai ([UTC+05:30](https://en.wikipedia.org/wiki/UTC%2B05:30))
    "+14:00" for Kiribati ([UTC+14:00](https://en.wikipedia.org/wiki/UTC%2B14:00))

I hope this answers.

dinogun: I am thinking that we need a short para in the doc MD on the timestamp format, so that the users know exactly what is expected. Please add that as part of this PR

dinogun (approved):
LGTM

PR #563, opened 2023-02-10
Set AUTOTUNE_PORT based on environment variable

patilsuraj767:


dinogun (approved):
Approving as it doesn't change the default behavior and there is no reason to be tied only to port `8080`

PR #565, opened 2023-02-15
Adds cluster name in kruize object

bharathappali:
This PR introduces the `cluster_name` attribute to Kruize Object.

signed-off: bharathappali <abharath@redhat.com>

### DIFF
--- a/src/main/java/com/autotune/analyzer/deployment/KruizeDeployment.java
+++ b/src/main/java/com/autotune/analyzer/deployment/KruizeDeployment.java
@@ -419,7 +419,8 @@ public class KruizeDeployment {
                     targetCluster,
                     sloInfo,
                     selectorInfo,
-                    objectReference
+                    objectReference,
+                    null
             );
 
         } catch (InvalidValueException | NullPointerException | JSONException e) {
--- a/src/main/java/com/autotune/analyzer/deployment/KruizeDeployment.java
+++ b/src/main/java/com/autotune/analyzer/deployment/KruizeDeployment.java
@@ -326,6 +326,7 @@ public class KruizeDeployment {
             String name;
             String mode;
             String targetCluster;
+            String clusterName;
             SloInfo sloInfo;
             ObjectiveFunction objectiveFunction = null;
             String namespace;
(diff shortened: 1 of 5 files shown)


src/main/java/com/autotune/common/k8sObjects/KruizeObject.java
```diff
@@ -54,14 +54,18 @@ public final class KruizeObject {
...
 
     public KruizeObject(String experimentName,
                         String namespace,
                         String mode,
                         String targetCluster,
                         SloInfo sloInfo,
                         SelectorInfo selectorInfo,
-                        ObjectReference objectReference) throws InvalidValueException {
+                        ObjectReference objectReference,
+                        String clusterNameContent
```
dinogun, line 67: Move this parameter to immediately after `experimentName` and it should be called `clusterName`

src/main/java/com/autotune/common/k8sObjects/KruizeObject.java
```diff
@@ -226,8 +232,17 @@ public void setValidationData(ValidationResultData validationData) {
...
+        return clusterName;
+    }
+
     @Override
     public String toString() {
+        // Creating a temparory cluster name as we allow null for cluster name now
+        // Please change it to use `clusterName` variable itself if there is a null check already in place for that
+        String tmpClusterName = "";
+        if (clusterName != null)
+            tmpClusterName = clusterName;
```
dinogun, line 245: Why is this a special case? `clusterName` = "null" should be handled by the same regular code itself

src/main/java/com/autotune/analyzer/deployment/KruizeDeployment.java
```diff
@@ -419,7 +419,8 @@ private static KruizeObject getAutotuneObject(String autotuneObjectJsonStr) {
                     targetCluster,
                     sloInfo,
                     selectorInfo,
-                    objectReference
+                    objectReference,
+                    null
```
dinogun, line 423: Why is this being set to null? We need to read in an `optstring` which will set to empty string if the user has not set it

src/main/java/com/autotune/common/k8sObjects/KruizeObject.java
bharathappali: Done.

src/main/java/com/autotune/common/k8sObjects/KruizeObject.java
bharathappali: Thanks for your review @dinogun  As there was no `null` check in place for `clusterName` I was substituting empty string if the cluster name is `null` to avoid null be printed as cluster name if kruize object is printed.

It also contradicts with a cluster named as `NULL` if we allow printing null as cluster name if it's not set

src/main/java/com/autotune/analyzer/deployment/KruizeDeployment.java
bharathappali: Ah! I missed that. I placed it for null testing, updated the code to substitute what user has sent.

dinogun:
@bharathappali The API.md changes are missing, please update the documentation

bharathappali:
@dinogun I have raised a PR #569 to add doc and test changes for the cluster name. Will be covering the API.md changes in that PR if you are comfortable.

dinogun (approved):
LGTM

PR #587, opened 2023-03-02
Fix cors issue for UI to access the API's

bharathappali:
@dinogun Currently the UI is not able to access these API's as UI and kruize are running on different domains. This is a temporary change as we need to move towards a web filter model once we finalise API scope (which API are internal and which are client facing or external).

I have added todos mentioning this current patch should not be used after implementing a web filter.

I'm currently making it draft as I would like to get feedback from bhanvi upon testing it with front end.


bhanvimenghani:
@bharathappali Thanks for this, the apis are now working fine, no cors error for Namespace, Deployments, ListPerformanceProfiles & ListExperiments.

bharathappali:
Thanks for confirming @bhanvimenghani Will make PR review ready

dinogun (approved):
LGTM

PR #588, opened 2023-03-02
Fix recommendations key

bharathappali:
@dinogun The key is changed to `recommendations` and reduced the api output in doc


src/main/java/com/autotune/common/k8sObjects/ContainerObject.java
```diff
@@ -30,6 +30,7 @@ public class ContainerObject {
     private String image;
     private String container_name;
     private HashMap<Timestamp, StartEndTimeStampResults> results;
+    @SerializedName(AutotuneConstants.JSONKeys.RECOMMENDATIONS)
     private HashMap<Timestamp, HashMap<String, Recommendation>> recommendation;
```
dinogun, line 34: I think we need to modify the variable name as well to `recommendations` as it stores multiple ones anyway

bharathappali: Done

dinogun (approved):
LGTM

PR #595, opened 2023-03-05
Adds service to list the supported k8s object types

bharathappali:
Adds a service to list the supported K8S object types by kruize

Resolves the first item mentioned in #591 (`One API to get k8s object corresponding to a namespace selected`)

### DIFF
--- a/src/main/java/com/autotune/analyzer/Analyzer.java
+++ b/src/main/java/com/autotune/analyzer/Analyzer.java
@@ -65,5 +65,6 @@ public class Analyzer {
         // Adding UI support API's
         context.addServlet(ListNamespaces.class, ServerContext.LIST_NAMESPACES);
         context.addServlet(ListDeploymentsInNamespace.class, ServerContext.LIST_DEPLOYMENTS);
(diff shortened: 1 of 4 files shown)


bharathappali:
@dinogun As part of features which needs to be added to fix #591 I have come up with a simple basic api to list out the supported k8s types by kruize.

Currently the API response looks something like this

```
{
  "kubernetes_objects": [
    "deployment",
    "deploymentConfig",
    "statefulset",
    "replicaset",
    "replicationController",
    "daemonset"
  ]
}
```

This is just a basic design of API and currently it doesn't cover any other attributes except the list of k8s objects. Please let me know if we need to add any other attributes for this API. 

Thanks in advance!

dinogun:
@bharathappali doesn't this need to add CORS support?

bharathappali:
Yes @dinogun CORS support needs to be added, We are having an issue opened to discuss about mechanism to add CORS support. Please add views on the issue #640 . Will update the PR based on the discussion outcome. Thanks in advance!

dinogun (approved):
LGTM

PR #607, opened 2023-03-06
Updated the scripts and jsons as per latest design changes

chandrams:
Updated the scripts and jsons as per latest design changes


dinogun (approved):
LGTM

PR #608, opened 2023-03-06
Fix deploy script for CRC case with custom docker image

dinogun:
Update sed to ignore the container image tag as that need not be the same as the pom version


chandrams (approved):
LGTM

PR #624, opened 2023-03-09
Better Cleanup on CRC termination

dinogun:


manifests/crc/default-db-included-installation/minikube/kruize-crc-minikube.yaml
```diff
@@ -1,9 +1,4 @@
 apiVersion: v1
-kind: Namespace
```
chandrams, line 2: @dinogun - Do we need a similar change in kruize-crc-openshift.yaml?

scripts/minikube-helpers.sh
```diff
@@ -189,14 +189,21 @@ function minikube_crc_start() {
 	echo "###   Installing kruize for minikube"
 	echo
 
-	MANIFEST_FILE=$KRUIZE_DEPLOY_MANIFEST_MINIKUBE
+	MANIFEST_FILE=${KRUIZE_CRC_DEPLOY_MANIFEST_MINIKUBE}
 	autotune_ns="monitoring"
```
chandrams, line 193: Please add the below check here as well
```
if [ -z "$autotune_ns" ]; then
	autotune_ns="monitoring"
fi
```

msvinaykumar (approved):

manifests/crc/default-db-included-installation/minikube/kruize-crc-minikube.yaml
dinogun: No, in openshift we own the `openshift-tuning` ns, so we can create and delete it when we start and terminate kruize respectively. Whereas in minikube `monitoring` already exists and cannot be deleted

scripts/minikube-helpers.sh
dinogun: done

chandrams (approved):
LGTM

PR #644, opened 2023-03-15
Add CORS filter in the servlet context

bharathappali:


src/main/java/com/autotune/utils/AutotuneConstants.java
```diff
@@ -307,4 +307,17 @@ private ListDeploymentsInNamespace() { }
             }
         }
     }
+
+    public static class CORSConstants {
```
dinogun, line 311: javadoc here

src/main/java/com/autotune/utils/filter/KruizeCORSFilter.java
```diff
@@ -0,0 +1,56 @@
...
+package com.autotune.utils.filter;
+
+import com.autotune.utils.AutotuneConstants;
+import org.eclipse.jetty.servlet.FilterHolder;
+import org.eclipse.jetty.servlets.CrossOriginFilter;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class KruizeCORSFilter {
```
dinogun, line 26: javadoc

dinogun:
Add a comment in the `ServerContext` to indicate that all end point have CORS enabled by default.
https://github.com/kruize/autotune/blob/2263afaea33b3b96f21343db9e423b9d23e52505/src/main/java/com/autotune/utils/ServerContext.java#L27

src/main/java/com/autotune/utils/AutotuneConstants.java
bharathappali: Thanks for the review @dinogun , Added the javadoc.

PR #674, opened 2023-03-23
db tables created and stored for create and update results api.

msvinaykumar:


src/main/java/com/autotune/database/table/ExperimentDetailTable.java
```diff
@@ -0,0 +1,135 @@
...
+ * targetCluster: A string representing the target cluster for the experiment.
+ * performance_profile: A string representing the performance profile for the experiment.
+ * status: An enum representing the status of the experiment, defined in AnalyzerConstants.ExperimentStatus.
+ * extended_data: A JSON object representing extended data for the experiment.
+ * meta_data: A string representing metadata for the experiment.
+ * The ExperimentDetail class also has getters and setters for all its fields.
+ */
+@Entity
+@Table(name = "experiment_detail")
+public class ExperimentDetailTable {
```
dinogun, line 44: This is just a table entry, not the table itself, so it needs to be named more appropriately (Eg. KruizeExperimentEntry). Also the table name should be `kruize_experiment`

src/main/java/com/autotune/service/InitiateListener.java
```diff
@@ -108,7 +108,7 @@ public void run() {
                     }
             );
         };
-        createExperimentExecutorScheduled.scheduleAtFixedRate(checkForNewExperiment, 5, 5, TimeUnit.SECONDS);
+        createExperimentExecutorScheduled.scheduleAtFixedRate(checkForNewExperiment, 1, 1, TimeUnit.SECONDS);
```
dinogun, line 111: Can we have these defined as constants

src/main/java/com/autotune/analyzer/utils/ServiceHelpers.java
```diff
@@ -396,6 +404,81 @@ public static ExperimentResultData convertUpdateResultsSOToExperimentResultData(
                 experimentResultData.setDeployments(deploymentResultDataList);
                 return experimentResultData;
             }
+
+            /**
+             * @param kruizeObject
+             * @return ExperimentDetail
+             * This methode facilitate to store data into db by accumulating required data from KruizeObject.
+             */
+            public static ExperimentDetailTable convertKruizeObjectToExperimentDBObj(KruizeObject kruizeObject) {
```
dinogun, line 413: This needs to be in a `DBHelpers` class inside the `database` package

src/main/java/com/autotune/analyzer/experiment/ExperimentInterfaceImpl.java
```diff
@@ -50,9 +57,27 @@ public boolean addExperimentToLocalStorage(Map<String, KruizeObject> mainKruizeE
 
     @Override
     public boolean addExperimentToDB(KruizeObject kruizeObject) {
```
dinogun, line 59: Looking at these functions, I think these need to be in a `ExperimentDBInterface` inside the `database` package. We can then have a `postgresDBImpl` that can host all of these functions.

src/main/java/com/autotune/analyzer/experiment/ExperimentInterfaceImpl.java
```diff
@@ -50,9 +57,27 @@ public boolean addExperimentToLocalStorage(Map<String, KruizeObject> mainKruizeE
...
-        //TODO insert in to db
+        boolean success = false;
         updateExperimentStatus(kruizeObject, AnalyzerConstants.ExperimentStatus.IN_PROGRESS);
-        return true;
+        ExperimentDetailTable experimentDetailTable = ServiceHelpers.Converters.KruizeObjectConverters.convertKruizeObjectToExperimentDBObj(kruizeObject);
```
dinogun, line 62: Where are we doing the error checking for this and how are the errors captured for the user to be made aware of?

src/main/java/com/autotune/database/table/ExperimentResultsTable.java
```diff
@@ -0,0 +1,74 @@
...
+import jakarta.persistence.*;
+import org.hibernate.annotations.JdbcTypeCode;
+import org.hibernate.type.SqlTypes;
+
+import java.sql.Timestamp;
+
+@Entity
+@Table(name = "experiment_results",
+        uniqueConstraints = {@UniqueConstraint(columnNames = {"experiment_name", "start_timestamp", "end_timestamp"})})
+public class ExperimentResultsTable {
```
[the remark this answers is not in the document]
msvinaykumar: We have named one of our tables as "kruize_experiment", with an underscore '_' separating the words. On the other hand, we have named another table "KruizeResultsEntry" without any underscore between words.

src/main/java/com/autotune/analyzer/experiment/ExperimentInterfaceImpl.java
msvinaykumar: @dinogun As we mentioned in our previous call, the insertion of data into the database occurs only after the message is added to the queue. In order to provide a report to the user, we would need to implement a UUID mechanism. Currently, if there are any errors during the database insertion process, the only way to identify them is through the log. However, we can temporarily resolve this issue until the UUID system is implemented by eliminating the queue for single uploads and performing the database insertion immediately after validation. Please inform me if this solution is necessary for the 12th release.

dinogun: The queue was to be implemented only for the case when multi record arrays are fed to createExperiment/updateResults. For single records, we can insert into DB directly after completing validation checking for both createExperiment and updateResults. This is what is needed for `0.0.12_rm` release

src/main/java/com/autotune/database/dao/ExperimentDAOImpl.java
```diff
@@ -0,0 +1,89 @@
...
+            } catch (HibernateException e) {
+                LOGGER.error("Not able to save experiment due to {}", e.getMessage());
+                updateExperimentStatus(kruizeObject, AnalyzerConstants.ExperimentStatus.FAILED);
+                if (tx != null) tx.rollback();
+                e.printStackTrace();
+                validationOutputData.setSuccess(false);
+                validationOutputData.setMessage(e.getMessage());
+                //todo save error to API_ERROR_LOG
+            }
+        }
```
dinogun, line 45: catch is missing

msvinaykumar: Using Try with Resources.

src/main/java/com/autotune/database/table/KruizeRecommendationEntry.java
```diff
@@ -0,0 +1,54 @@
...
+
+import com.fasterxml.jackson.databind.JsonNode;
+import jakarta.persistence.*;
+import org.hibernate.annotations.JdbcTypeCode;
+import org.hibernate.type.SqlTypes;
+
+import java.sql.Timestamp;
+
+@Entity
+@Table(name = "kruize_recommendation")
```
dinogun, line 11: This needs to be `kruize_recommendations` as per the doc

src/main/java/com/autotune/database/table/KruizeExperimentEntry.java
```diff
@@ -0,0 +1,154 @@
...
+ * mode: A string representing the mode of the experiment.
+ * targetCluster: A string representing the target cluster for the experiment.
+ * performance_profile: A string representing the performance profile for the experiment.
+ * status: An enum representing the status of the experiment, defined in AnalyzerConstants.ExperimentStatus.
+ * extended_data: A JSON object representing extended data for the experiment.
+ * meta_data: A string representing metadata for the experiment.
+ * The ExperimentDetail class also has getters and setters for all its fields.
+ */
+@Entity
+@Table(name = "kruize_experiment")
```
dinogun, line 44: This needs to be `kruize_experiments` as per the doc

dinogun (approved):
LGTM. Thanks for all the effort to get this done on time @msvinaykumar!

PR #691, opened 2023-03-30
Livenessprobe script changes.

msvinaykumar:
The liveness probe script has been modified such that the session factory object is now created using a singleton within the Kruize context. This is a departure from the earlier setup, where the session factory object was recreated every time and was running outside the Kruize context.


dinogun:
Closing stale PRs

PR #692, opened 2023-03-30
Change DB config precedence

### LINKED ISSUE #685: [Database] Environment variable should took precedence over configuration file.
Here - https://github.com/kruize/autotune/blob/master/src/main/java/com/autotune/database/init/KruizeHibernateUtil.java#L42-L62 I can see kruize first checks for configuration file and if file not found then looks for the env variable.
Request is to first get database config from configuration files and then override the config if any database environment variable is set.

khansaad:
This PR fixes #685


dinogun (approved):
LGTM

PR #699, opened 2023-04-03
Add note in docs to use timeseries order for update results API

bharathappali:
Currently autotune supports the results to be updated in a timeseries order. Adding a note for the same in the documentation.

@dinogun Can I please have your review?


dinogun:
@patilsuraj767 We looked into out of order posting of `updateResults` for an experiment. It looks like this will cause an exponential recalculation of `recommendations` for each recommendation that was generated after the `interval_end_time` of the current `updateResults`. So we have decided to not support that for the `single` updateResults case. In the future, when we have bulk APIs implemented, it can be supported where the individual entries of a bulk posting can be out of order.

patilsuraj767:
Ok, Thanks for the confirmation @dinogun and @bharathappali

PR #713, opened 2023-04-04
Remove result data from Kruize Object

### LINKED ISSUE #700: Remove resultData from the KruizeObject
Currently we're a getting a redundant result response in the `listExperiments` API. We need to remove the `resultData` object from the JSON and keep the results inside `kubernetes_objects`

khansaad:
This PR fixes #700


dinogun (approved):
LGTM

PR #717, opened 2023-04-05
Layer presence label value validation

ozer550:
Added Validation For layer_presence_label_value (WIP)

Changes:
The PR is assuming that the validation for the label_value being not present or NULL cannot be done in CRD validation as either one of the keys i.e (presence/queries/label should be present and this type of validation is not possible in CRD itself).

Concerns:
For all the three yaml test cases:
no-layer-presence-labelvalue
no-layer-presence-labelvalue-value
null-layer-presence-labelvalue

Value returned for extracting the string "layer_presence_label_value" comes out to be blank or empty. Which is not expected logically. The value string never comes out to be null but blank for every negative test cases.

@dinogun Can you please guide on what should be considered valid and invalid for the invalid-layer-presence-labelvalue case?


dinogun:
Closing stale PRs

PR #718, opened 2023-04-06
all Env vars moved to kube config object.

msvinaykumar:


src/main/java/com/autotune/operator/InitializeDeployment.java
```diff
@@ -15,66 +15,131 @@
...
+                    }
+                    envValue = primaryObj.get(parts[i]);
+                    message = message + "  :- found in mounted config file";
+                } catch (Exception e) {
+                    message = message + "  :- Not found in mounted config file";
+                }
+            } else {
+                message = message + "  :- No mount config file";
+            }
+            //Override if env value set outside kruizeConfigJson
```
dinogun, line 121: Has this been tested? (ENV vars getting precedence over config JSON?)

src/main/java/com/autotune/operator/KruizeDeploymentInfo.java
```diff
@@ -38,20 +37,36 @@
...
+    public static String SETTINGS_HIBERNATE_HBM2DDL_AUTO;
+    public static String SETTINGS_HIBERNATE_SHOW_SQL;
+    public static String AUTOTUNE_MODE;
+    public static String MONITORING_AGENT;
+    public static String MONITORING_SERVICE;
+    public static String MONITORING_AGENT_ENDPOINT;
+    public static String CLUSTER_TYPE;
+    public static String K8S_TYPE;
+    public static String AUTH_TYPE;
+    public static String AUTH_TOKEN;
```
dinogun, line 65: Do these vars now have some special significance? It looks like they are also being over loaded as ENV params?

src/main/java/com/autotune/operator/InitializeDeployment.java
msvinaykumar: Certainly. Kruize follows a specific logic when it comes to configuration variables. If a configuration variable is set but not defined in the environment variables, Kruize will use the configuration variables. If a configuration variable is not set, but it is defined in the environment variables, Kruize will use the environment variables. However, if both configuration variables and environment variables are set, Kruize will prioritize the environment variables and override the configuration variables.

src/main/java/com/autotune/operator/KruizeDeploymentInfo.java
msvinaykumar: yes this values can be set either via config json or via ENV.

dinogun (approved):
LGTM

PR #721, opened 2023-04-06
Fix timestamp naming

dinogun:
Looks like the start_timestamp, end_timestamp -> interval_start_time, interval_end_time was missed out for two more classes

### DIFF
--- a/src/main/java/com/autotune/common/data/result/ExperimentResultData.java
+++ b/src/main/java/com/autotune/common/data/result/ExperimentResultData.java
@@ -31,9 +31,9 @@ public class ExperimentResultData {
     private String experiment_name;
     private String trialNumber;
     @SerializedName("interval_start_time")
-    private Timestamp starttimestamp;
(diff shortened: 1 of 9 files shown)


dinogun:
@msvinaykumar please make sure this does not have implications from a DB point of view

bharathappali (approved):
LGTM!

src/main/java/com/autotune/common/data/result/ExperimentResultData.java
```diff
@@ -31,9 +31,9 @@ public class ExperimentResultData {
     private String experiment_name;
     private String trialNumber;
     @SerializedName("interval_start_time")
```
khansaad, line 33: I think this also can be changed to it's corresponding constant value similar to how it's done in the IntervalResults class

dinogun: done, thanks

PR #750, opened 2023-04-17
Fix updateResults duration check

dinogun:


dinogun:
@chandrams This should fix the duration check for updateResults in tests. Please check

chandrams:
@dinogun - Thanks for the fix. Tested it, a small update is required to the below message, 5 seconds needs to 30 seconds.

"In the message "Interval duration cannot be less than or greater than measurement_duration by more than 5 seconds"

dinogun:
@chandrams sorry yeah I missed that, added it now

chandrams (approved):
LGTM

PR #760, opened 2023-04-19
Moved capturing ffdc logs into a separate step

chandrams:
Moved capturing ffdc logs into a separate step as it is not executed when the deploy script fails


dinogun (approved):
LGTM

PR #761, opened 2023-04-20
Remove watcher exception from log.

msvinaykumar:
and just print
"Watcher not added! Only REST API access is enabled."


dinogun (approved):
LGTM

PR #766, opened 2023-04-21
Developer quick handy tool to test kruize Scalability.

msvinaykumar:
This is a handy tool for developers to check scalability. By running the "quickTestScalability.py" script with the specified parameters,

python3 quickTestScalability.py --ip master-1.kruizevin.lab.psi.pnq2.redhat.com --port 31620 --name firstEXP --count 2,10 --minutesjump=15

above script creates default performance profile
followed by  two experiments with the names "firstEXP_1" and "firstEXP_2" are generated.
Each experiment will have 10 results, 
and the start and end times for each result will be auto-incremented by 15 minutes.

### DIFF
--- a/tests/scripts/quickTestScalability.py
+++ b/tests/scripts/quickTestScalability.py
@@ -335,36 +335,35 @@ for i in range(expcount):
         # Check the response
         if response.status_code == 201:
             print('Request successful!')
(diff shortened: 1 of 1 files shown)


dinogun (approved):
Approving so that we can capture all the improvements

PR #768, opened 2023-04-22
Set Kruize CPU and Mem request and Java Heap Size

dinogun:
Set Kruize Requests and Limits as follows (OpenShift only)
- CPU = 4 cores
- Mem = 4Gi
- Java Heap = -XX:MaxRAMPercentage=80  
```
# This means  Heap Size = 80% 4Gi = 3.2Gi
# Native mem = 819Mi
```
Not setting any config for minikube as it is needed to be run in constrained environments


chandrams (approved):
LGTM

PR #771, opened 2023-04-22
Updated the workflows to capture ffdc logs always irrespective of the status of previous step

chandrams:
Updated the workflows to capture ffdc logs always irrespective of the status of previous step


dinogun (approved):
LGTM

PR #772, opened 2023-04-22
added system.gc.

msvinaykumar:
Calling System GC for every 100 success results.


dinogun (approved):
Approving as a temp measure for now

PR #777, opened 2023-04-24
Fix variation for limits

bharathappali:
The current limits are set same in variation due to variable change. This PR sets the correct variables to be placed.


dinogun (approved):
LGTM

PR #784, opened 2023-04-27
Added ssl configuration .

msvinaykumar:
Added relevant ssl enabled configuration which can be set either via clowder or by using env


dinogun:
Closing stale PRs

PR #792, opened 2023-04-28
Use DB from servlets instead of map

dinogun:


src/main/java/com/autotune/database/service/ExperimentDBService.java
```diff
@@ -169,6 +169,169 @@ public void loadAllExperimentsData() throws Exception {
         LOGGER.debug(KruizeOperator.autotuneObjectMap.toString());
     }
 
+    public void loadExperimentByName(Map<String, KruizeObject> mainKruizeExperimentMap, String experimentName) throws Exception {
```
bharathappali, line 172: I feel all the functions in this class are independent and can be `static` functions as the operations are being done on the objects passed to function and no class level changes are observed.

If we need to reduce the use or static area, we can make the `ExperimentDBService` as a singleton and jus maintain only one object to call all these functions

These changes need not be done in this PR, Just recording my observation here.

dinogun:
This PR is currently being tested for scalability with the following options
CPU (Request, Limit) = (3, None)
Mem (Request, Limit) = (4Gi, 5Gi)

10,000 experiments being updated for results in parallel with 1 day increments.
So far the observations have been
Max CPU = 4.7 cores
Max Mem = 2.18GiB

msvinaykumar (approved):
LGTM

PR #795, opened 2023-05-01
Fix variation object

bharathappali:
The current implementation of variation needs to be like

variation_object

But we have changed the `currentCPULimits` to `currentCPURequest` and `currentMemLimits` to `currentMemRequest` at the time of setting the limit values of recommendation same as requests. This causes issue like showing Limits and Request not set notifications to appear in the case only where request are set but limits exist. And some times this allows empty variation object to be added.

This PR checks if map is not empty before adding and also extracts the current limits and requests for calculating variation


bharathappali:
@dinogun Can I please have your reviews on this PR?

dinogun (approved):
LGTM

PR #800, opened 2023-05-02
Set Allow private network to true

bharathappali:
Kruize UI is facing issues with private network

```
Access to fetch at 'http://192.168.39.141:30487/listRecommendations?experiment_name=quarkus-resteasy-kruize-min-http-response-time-db_0&latest=false' from origin 'http://0.0.0.0:8080/' has been blocked by CORS policy: The request client is not a secure context and the resource is in more-private address space `private`.
```

Added a response header `Access-Control-Allow-Private-Network` and setting it to `true`


bharathappali:
@dinogun Can I please have your review on this?

bharathappali:
closing this PR as the NGINX will solve the problem as the call will happen from same IP

PR #811, opened 2023-05-08
Fix recommendation issue: CPU Recommendation cannot be zero or negative. Test Changes Included

bharathappali:
Users are seeing zero recommendations in remote monitoring usecase if the CPU is idle (CPU Usage = 0). As Recommendation should not suggest `kill your workload`, We need to have a min recommendation values for CPU and Memory.

At start we can have some hardcoded fixed values which are generic, and later on move to previous data based min CPU as recommendation if CPU is idle for the recommendation duration.

@dinogun Please lemme know your views on this


src/main/java/com/autotune/analyzer/recommendations/engine/DurationBasedRecommendationEngine.java
```diff
@@ -379,6 +379,10 @@ private static RecommendationConfigItem getCPURequestRecommendation(Map<Timestam
             cpuRequest = CommonUtils.percentile(NINETY_EIGHTH_PERCENTILE, cpuUsageList);
         }
 
+        // Make sure that the recommendation cannot be 0 or -ve
```
dinogun, line 382: If the cpu usage is zero at this point, that means that the usage for atleast 24 hrs (short term) was zero. I'd say that we should create a notification at this point that the (appropriate) term has not enough cpu data to arrive at a recommendation.

Also, depending on which env the workload in running, `0.1` can be too much! (Eg embedded), so not good to fix that. Unless there is no history of any usage, what this means is that the workload was idling during this observation period and that should just be a notification saying the same

bharathappali: @dinogun This gets me a thought on the flow of the recommendations, I was in an impression that if we have data with us we need to provide a recommendation which is satisfactory, So every result after a certain threshold (min data) we generate a recommendation. Here are my views on how we see this

**Assumption-A: Every result after a certain min data requirement should have a recommendation**

In this type of assumption, we check if the recommendation is non zero, if it's zero then we lookup for last applied cpuRequest and recommend the same, if there is no cpuRequest set we return min recommendation hardcoded.

**Assumption-B: Recommendation is provided only if some constraints are met even after min data requirement**

In this type of assumption, we check if the recommendation is non zero, if it's zero we add a notification saying that recommendation cannot be generated as the workload is idle

I was thinking if we extend assumption A a bit further, we can detect the idle entries (cpu usage = 0) and avoid them for recommendation calculation and calculate recommendation on only non-idle data, This makes us consider the 0 cpuUsage entries as missing data and we will not consider that entries duration in minutes at all, So this keeps it straight forward that if there is min valid and non - zero data for a term (eg. 24 hrs) exists before we give our first recommendation, else we by default add not enough data for recommendation notification.

I'm interested to know in which way we can go forward Assumption A or B. I couldn't decide upon one way as each of them has their own share of pros and cons. Please lemme know your views on this.

PS: I have added my views (in Assumption A) assuming that having notification after min data is mandatory, if my base assumption itself is wrong (please correct me if it is) then we can proceed with Assumption B.

Thanks in advance!

kusumachalasani: @bharathappali  By generating recommendations based on available data alone, both relationship building and advanced analysis can be achieved. For example, identifying consumption patterns for weekday,  specific events or holidays.
Notifications can also inform users about these behavioral changes, and if needed, they can consider previous or short-term recommendations.

In cases where a memory-bound pod is present, CPU consumption may drop to zero while memory usage remains high. So, it is important not to ignore idle CPU usage.
So, it would be better to consider the data available during that duration alone and provide recommendations or notifications.

bharathappali:
@dinogun the PR is ready for review, Can I please have your views on these changes?

Changes:
- Zero CPU usage will return a notification

bharathappali:
Example output of listRecommendations:

```
[
    {
        "cluster_name": "cluster-one-division-bell",
        "kubernetes_objects": [
            {
                "type": "deployment",
                "name": "my-sample-deployment",
                "namespace": "default",
                "containers": [
                    {
                        "container_image_name": "my-sample-image",
                        "container_name": "my-sample-container",
                        "recommendations": {
                            "notifications": {
                                "1001": {
                                    "type": "info",
                                    "message": "Duration Based Recommendations Available",
                                    "code": 1001
                                }
                            },
                            "data": {
                                "2023-05-18T01:11:15.061Z": {
                                    "duration_based": {
                                        "short_term": {
                                            "monitoring_start_time": "2023-05-17T01:11:15.061Z",
                                            "monitoring_end_time": "2023-05-18T01:11:15.061Z",
                                            "duration_in_hours": 24.0,
                                            "pods_count": 0,
                                            "confidence_level": 0.0,
                                            "config": {
                                                "limits": {
                                                    "memory": {
                                                        "amount": 1178.556,
                                                        "format": "MiB"
                                                    }
                                                },
                                                "requests": {
                                                    "memory": {
                                                        "amount": 1178.556,
                                                        "format": "MiB"
                                                    }
                                                }
                                            },
                                            "variation": {
                                                "limits": {
                                                    "memory": {
                                                        "amount": 950.4060000000001,
                                                        "format": "MiB"
                                                    }
                                                },
                                                "requests": {
                                                    "memory": {
                                                        "amount": 738.426,
                                                        "format": "MiB"
                                                    }
                                                }
                                            },
                                            "notifications": {
                                                "1003": {
                                                    "type": "info",
                                                    "message": "CPU usage is zero, No CPU Recommendations can be generated",
                                                    "code": 1003
                                                }
                                            }
                                        },
                                        "medium_term": {
                                            "pods_count": 0,
                                            "confidence_level": 0.0,
                                            "notifications": {
                                                "1000": {
                                                    "type": "info",
                                                    "message": "There is not enough data available to generate a recommendation.",
                                                    "code": 1000
                                                }
                                            }
                                        },
                                        "long_term": {
                                            "pods_count": 0,
                                            "confidence_level": 0.0,
                                            "notifications": {
                                                "1000": {
                                                    "type": "info",
                                                    "message": "There is not enough data available to generate a recommendation.",
                                                    "code": 1000
                                                }
                                            }
                                        }
                                    }
                                }
                            }
                        }
                    }
                ]
            }
        ],
        "version": "1.0",
        "experiment_name": "sensational-kusuma"
    }
]
```

src/main/java/com/autotune/analyzer/recommendations/engine/DurationBasedRecommendationEngine.java
```diff
@@ -92,61 +93,90 @@ public HashMap<String, Recommendation> generateRecommendation(ContainerData cont
...
-                if (null != cpuRequestItem) {
+                if (null != cpuRequestItem && isCpuRequestValid) {
                     generatedCpuRequest = cpuRequestItem.getAmount();
                     generatedCpuRequestFormat = cpuRequestItem.getFormat();
                     requestsMap.put(AnalyzerConstants.RecommendationItem.cpu, cpuRequestItem);
+                } else {
+                    RecommendationNotification recommendationNotification = new RecommendationNotification(
+                            AnalyzerConstants.RecommendationNotification.CPU_RECORDS_ARE_ZERO
+                    );
+                    recommendation.addNotification(recommendationNotification);
```
dinogun, line 121: This should happen in getCPURequestRecommendation method itself. So we might need to send the `recommendationNotification` object as a parameter. This will help catch other scenarios which can only be caught inside the method.

src/main/java/com/autotune/analyzer/recommendations/engine/DurationBasedRecommendationEngine.java
```diff
@@ -92,61 +93,90 @@ public HashMap<String, Recommendation> generateRecommendation(ContainerData cont
...
-                if (null != memRequestItem) {
+                if (null != memRequestItem && isMemoryRequestValid) {
                     generatedMemRequest = memRequestItem.getAmount();
                     generatedMemRequestFormat = memRequestItem.getFormat();
                     requestsMap.put(AnalyzerConstants.RecommendationItem.memory, memRequestItem);
+                } else {
+                    RecommendationNotification recommendationNotification = new RecommendationNotification(
+                            AnalyzerConstants.RecommendationNotification.MEMORY_RECORDS_ARE_ZERO
+                    );
+                    recommendation.addNotification(recommendationNotification);
```
dinogun, line 133: Same as above

src/main/java/com/autotune/analyzer/recommendations/engine/DurationBasedRecommendationEngine.java
bharathappali: @dinogun Thanks for the review, I really appreciate your idea of passing notification to the function as param, But  after going through the code I have came up with points to be discussed before I could make some changes as suggested, Please correct me if I'm wrong.

- I'm more inclined keep the math calculation away from making structural changes to a data structure (Notification Object) whose scope is at a higher level
- As per your suggestion we can implement the notification updates at the calculation phase itself by implementing something like this

Generate recommendation calls `processCPUrecommendation` which returns a object with `requests` and `limits` map,  and a notification map

 `processCPUrecommendation` calls `checkCPURequest` (basically we call getcpurequest and based on output we set enum) which returns  enum which decribes the status (tells about issue clearly or states that there is no issue) and the value of cpurequest, if there is no issue we read the cpurequest value and format the `requests` map and if the enum shows issue we add the notification to map and return them to `generate recomemndation` and it reads the structures as it is and populate the recommendation object
 
 
 Please let me know your views on this

dinogun: I'd prefer not to do any processing of the data at the `generateRecommendation` level, instead they all need to happen inside `getXYZRequestRecommendation`

design/NotificationCodes.md
```diff
@@ -0,0 +1,75 @@
...
+| 112101   |  INFO   | DURATION_BASED_RECOMMENDATIONS_AVAILABLE | Specifies that the Duration Based Recommendations are available |       Duration Based Recommendations Available        |
+| 120001   |  INFO   |       NOT_ENOUGH_DATA      |       Specifies that required data is not enough to create a recommendation     |  There is not enough data available to generate a recommendation. |
+| 313001   | NOTICE |      CPU_RECORDS_ARE_IDLE        |       Specifies that the CPU recordings are IDLE (Not much active CPU usage is identified)      |  CPU usage is mostly idle, No CPU Recommendation can be generated |
+| 323001   | NOTICE |      CPU_RECORDS_ARE_ZERO       |       Specifies that the CPU recordings are ZERO      |  CPU usage is zero, No CPU Recommendations can be generated |
+| 523001   | CRITICAL |     CPU_REQUEST_NOT_SET       |      Specifies that the CPU Requests are not set for the pod       |  CPU Request Not Set |
+| 523002   | CRITICAL |     CPU_LIMIT_NOT_SET       |     Specifies that the CPU Limits are not set for the pod        |  CPU Limit Not Set |
```
dinogun, line 69: This needs to be a WARNING as discussed

src/main/java/com/autotune/analyzer/utils/AnalyzerConstants.java
```diff
@@ -199,9 +202,10 @@ public RecommendationSubCategory[] getRecommendationSubCategories() {
 
     public enum RecommendationNotificationTypes {
         INFO("info", 1),
-        WARN("warning", 2),
-        ERROR("error", 3),
-        CRITICAL("critical", 4);
+        NOTICE("notice", 2),
+        WARN("warning", 3),
+        ERROR("error", 4),
+        CRITICAL("critical", 5);
```
dinogun, line 208: Order should be `INFO, ERROR, NOTICE, WARNING, CRITICAL`

src/main/java/com/autotune/analyzer/utils/AnalyzerConstants.java
```diff
@@ -220,6 +224,494 @@ public int getSeverity() {
...
+        ),
+        MEMORY_REQUEST_NOT_SET (
+                NotificationCodes.MEMORY_REQUEST_NOT_SET,
+                RecommendationNotificationMsgConstant.MEMORY_REQUEST_NOT_SET,
+                RecommendationNotificationTypes.CRITICAL
+        ),
+        CPU_LIMIT_NOT_SET (
+                NotificationCodes.CPU_LIMIT_NOT_SET,
+                RecommendationNotificationMsgConstant.CPU_LIMIT_NOT_SET,
+                RecommendationNotificationTypes.CRITICAL
```
dinogun, line 271: Should be WARNING

src/main/java/com/autotune/analyzer/utils/AnalyzerConstants.java
```diff
@@ -220,6 +224,494 @@ public int getSeverity() {
...
+        public static final int INFO_GENERAL_NETWORK_START = 115000;
+        public static final int INFO_GENERAL_NETWORK_END = 115999;
+        public static final int INFO_GENERAL_DISK_START = 116000;
+        public static final int INFO_GENERAL_DISK_END = 116999;
+        public static final int INFO_GENERAL_POWER_START = 117000;
+        public static final int INFO_GENERAL_POWER_END = 117999;
+        public static final int INFO_GENERAL_FUTURE_1_START = 118000;
+        public static final int INFO_GENERAL_FUTURE_1_END = 118999;
+        public static final int INFO_GENERAL_FUTURE_2_START = 119000;
+        public static final int INFO_GENERAL_FUTURE_2_END = 119999;
```
dinogun, line 355: Dont need these anymore, instead just use `RESERVED`

src/main/java/com/autotune/analyzer/utils/AnalyzerConstants.java
```diff
@@ -220,6 +224,494 @@ public int getSeverity() {
...
+
+        public static final int INFO_DATA_AVAILABILITY_GENERAL_START = 120000;
+        public static final int NOT_ENOUGH_DATA = 120001;
```
dinogun, line 358: Would be good to add the prefix `INFO` for consistency

src/main/java/com/autotune/analyzer/utils/AnalyzerConstants.java
```diff
@@ -220,6 +224,494 @@ public int getSeverity() {
...
+        //          SubSystem - Disk:       226000 - 226999 (10% of availability)
+        //          SubSystem - Power:      227000 - 227999 (10% of availability)
+
+        public static final int NOTICE_GENERAL_INFO_START = 210000;
+        public static final int NOTICE_GENERAL_INFO_END = 212999;
+        public static final int NOTICE_GENERAL_CPU_START = 213000;
+        public static final int CPU_RECORDS_ARE_IDLE = 213001;
+        public static final int NOTICE_GENERAL_CPU_END = 213999;
+        public static final int NOTICE_GENERAL_MEMORY_START = 214000;
+        public static final int MEMORY_RECORDS_ARE_IDLE = 214001;
```
dinogun, line 397: This needs to be removed

design/NotificationCodes.md
```diff
@@ -0,0 +1,75 @@
...
+
+
+# Detailed Codes
+
+|   CODE   |  TYPE   |     SHORT NOTATION      |                         DESCRIPTION                         |                          MESSAGE                            |
+|----------|---------|------------------------|-------------------------------------------------------------|------------------------------------------------------------|
+| 112101   |  INFO   | DURATION_BASED_RECOMMENDATIONS_AVAILABLE | Specifies that the Duration Based Recommendations are available |       Duration Based Recommendations Available        |
+| 120001   |  INFO   |       NOT_ENOUGH_DATA      |       Specifies that required data is not enough to create a recommendation     |  There is not enough data available to generate a recommendation. |
+| 313001   | NOTICE |      CPU_RECORDS_ARE_IDLE        |       Specifies that the CPU recordings are IDLE (Not much active CPU usage is identified)      |  CPU usage is mostly idle, No CPU Recommendation can be generated |
+| 323001   | NOTICE |      CPU_RECORDS_ARE_ZERO       |       Specifies that the CPU recordings are ZERO      |  CPU usage is zero, No CPU Recommendations can be generated |
```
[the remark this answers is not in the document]
bharathappali: Added `CPU_RECORDS_NOT_AVAILABLE` and `MEMORY_RECORDS_NOT_AVAILABLE`

design/NotificationCodes.md
```diff
@@ -0,0 +1,75 @@
...
+
+
+
+# Detailed Codes
+
+|   CODE   |  TYPE   |     SHORT NOTATION      |                         DESCRIPTION                         |                          MESSAGE                            |
+|----------|---------|------------------------|-------------------------------------------------------------|------------------------------------------------------------|
+| 112101   |  INFO   | DURATION_BASED_RECOMMENDATIONS_AVAILABLE | Specifies that the Duration Based Recommendations are available |       Duration Based Recommendations Available        |
+| 120001   |  INFO   |       NOT_ENOUGH_DATA      |       Specifies that required data is not enough to create a recommendation     |  There is not enough data available to generate a recommendation. |
+| 313001   | NOTICE |      CPU_RECORDS_ARE_IDLE        |       Specifies that the CPU recordings are IDLE (Not much active CPU usage is identified)      |  CPU usage is mostly idle, No CPU Recommendation can be generated |
```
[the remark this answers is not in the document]
bharathappali: Updated in code and also in the md file. Thanks!

src/main/java/com/autotune/analyzer/utils/AnalyzerConstants.java
bharathappali: Updated the order. Thanks!

src/main/java/com/autotune/analyzer/utils/AnalyzerConstants.java
bharathappali: Made changes accordingly

src/main/java/com/autotune/analyzer/utils/AnalyzerConstants.java
bharathappali: Moved the constants to a better var names which describe their sections and subsections and subsystems as well

src/main/java/com/autotune/analyzer/utils/AnalyzerConstants.java
bharathappali: Added the respective prefixs for each of them

src/main/java/com/autotune/analyzer/utils/AnalyzerConstants.java
```diff
@@ -220,6 +224,494 @@ public int getSeverity() {
...
+        //          SubSystem - CPU:        223000 - 223999 (10% of availability)
+        //          SubSystem - Memory:     224000 - 224999 (10% of availability)
+        //          SubSystem - Network:    225000 - 225999 (10% of availability)
+        //          SubSystem - Disk:       226000 - 226999 (10% of availability)
+        //          SubSystem - Power:      227000 - 227999 (10% of availability)
+
+        public static final int NOTICE_GENERAL_INFO_START = 210000;
+        public static final int NOTICE_GENERAL_INFO_END = 212999;
+        public static final int NOTICE_GENERAL_CPU_START = 213000;
+        public static final int CPU_RECORDS_ARE_IDLE = 213001;
```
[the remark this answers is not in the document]
bharathappali: Added the respective prefixs for each of them

src/main/java/com/autotune/analyzer/utils/AnalyzerConstants.java
bharathappali: Removed.

dinogun (approved):
LGTM

PR #813, opened 2023-05-09
Fixed the test status code for invalid exp name, clustername, removed quotes

chandrams:
Fixed the test status code for invalid exp name, cluster name, removed quotes for some of the keys


dinogun (approved):
LGTM

PR #827, opened 2023-05-22
Fix boundary condition issue mentioned in #826

### LINKED ISSUE #826: [Recommendation] Boundary condition issue with recommendation generation
Currently if kruize has exact 24 hrs of data it is not able to give out recommendations due to a boundary condition issue which was mentioned by @chandrams and @kusumachalasani . 

@msvinaykumar found that the issue is with the way Recommendation engine is getting the start time. 

Kruize is having the check for sum of duration minutes at the start  of iteration, so if the list has exact number of entries as it can't iterate more the check doesn't happen in the last iteration. Hence kruize is not generating the recommendation if it has exact entries for a duration.

bharathappali:
This PR adds the `sum check` after the increment itself to avoid the boundary condition where kruize was not able to generate the recommendation

Fixes #826


dinogun:
@bharathappali please update the test results. Also the PR check is failing, it looks like the `e2e` test needs to be updated to incorporate this fix?

dinogun (approved):
LGTM

dinogun:
Merging this as the PR check failure has been confirmed to be a testcase issue

PR #828, opened 2023-05-22
Add Performance profile table in the DB

### LINKED ISSUE #790: After Kruize pod restarts performance profile is not loaded from DB
If kruize is restarted, since currently perfProfile is not present in the DB now, it fails giving error Failed to create experiment: Performance Profile doesn't exist : resource-optimization-openshift

khansaad:
This PR contains changes to add new table `kruize_performance_profiles` and it's corresponding DAO and Model classes.

This will fix #790


chandrams:
@khansaad - Please update the fault tolerant test to check kruize restart works without having to create the performance profile again.

src/main/java/com/autotune/database/table/KruizePerformanceProfileEntry.java
```diff
@@ -0,0 +1,80 @@
...
+ * profile_version: A string representing the version of the performance profile.
+ * k8s_type: A string representing kubernetes type.
+ * SLO: A string representing the slo class, direction, Objective function and function variables.
+ */
+@Entity
+@Table(name = "kruize_performance_profiles")
+public class KruizePerformanceProfileEntry {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long performance_profile_id;
```
dinogun, line 40: What is this id and why is this the IDENTITY?
@msvinaykumar this is the same problem that we had with the results table. Please use `name` as the IDENTITY instead

khansaad: Done

dinogun (approved):
LGTM

PR #830, opened 2023-05-23
Update results optimized

msvinaykumar:
The update results API has been optimized to improve throughput while reducing the amount of memory required.


msvinaykumar:
changes merged into https://github.com/kruize/autotune/pull/855

PR #835, opened 2023-05-29
Add metrics match for PR check

### LINKED ISSUE #819: [Metrics] Enable PR checks to monitor perf parameters

kusumachalasani:
This fixes #819

### DIFF
--- a/tests/scripts/remote_monitoring_tests/helpers/utils.py
+++ b/tests/scripts/remote_monitoring_tests/helpers/utils.py
@@ -35,6 +35,33 @@ EXP_EXISTS_MSG = "Experiment name already exists: "
 INVALID_DEPLOYMENT_TYPE_MSG = "Invalid deployment type: xyz"
 INVALID_INTERVAL_DURATION_MSG = "Interval duration cannot be less than or greater than measurement_duration by more than 30 seconds"
 
+createExperimentCount_STATUS = "Metric Category: API, Metric Type: count, Name: createExperiment, Value: 10.0"
+createExperimentSum_STATUS = "Metric Category: API, Metric Type: sum, Name: createExperiment, Value: 0.228229311"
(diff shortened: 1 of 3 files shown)


tests/scripts/remote_monitoring_tests/helpers/utils.py
```diff
@@ -35,6 +35,33 @@
 INVALID_DEPLOYMENT_TYPE_MSG = "Invalid deployment type: xyz"
 INVALID_INTERVAL_DURATION_MSG = "Interval duration cannot be less than or greater than measurement_duration by more than 30 seconds"
 
+createExperimentCount_STATUS = "Metric Category: API, Metric Type: count, Name: createExperiment, Value: 10.0"
+createExperimentSum_STATUS = "Metric Category: API, Metric Type: sum, Name: createExperiment, Value: 0.228229311"
+listRecommendationsCount_STATUS = "Metric Category: API, Metric Type: count, Name: listRecommendations, Value: 11.0"
+listRecommendationsSum_STATUS = "Metric Category: API, Metric Type: sum, Name: listRecommendations, Value: 0.349422898"
```
chandrams, line 41: @kusumachalasani - I think count may not change but sum can change even for the same no. of exps / results. Have you run this multiple times to check this? Should we specify a range instead?

kusumachalasani: On my system, the sum remained the same.  I need to validate if it is same for PR systems.
Also, the sum values mentioned here needs to be updated after the first run that happens on this PR system considering the resources it has. Will wait for the Bharath fix to validate.

chandrams: You can create the github workflow and run it from your branch by commenting out the failing assertion.

dinogun:
Closing stale PRs

PR #840, opened 2023-06-01
Fix notification schema

bharathappali:
This PR aims to fix the test fail which is expecting the notifications to be an array and also fixes the recommendation notification examples in remote monitoring doc

@chandrams @dinogun Can I please have your views on this please?

Thanks in advance!


chandrams (approved):
LGTM

PR #843, opened 2023-06-02
Adds empty check for other containers if one container is populated

bharathappali:
This PR fixes the null issue seen when only one container data is populated and recommendations are checked for unpopulated containers

### DIFF
--- a/src/main/java/com/autotune/analyzer/recommendations/engine/DurationBasedRecommendationEngine.java
+++ b/src/main/java/com/autotune/analyzer/recommendations/engine/DurationBasedRecommendationEngine.java
@@ -276,6 +276,10 @@ public class DurationBasedRecommendationEngine implements KruizeRecommendationEn
 
     @Override
     public boolean checkIfMinDataAvailable(ContainerData containerData) {
+        // Check if data available
+        if (null == containerData || null == containerData.getResults() || containerData.getResults().isEmpty()) {
+            return false;
+        }
         // Initiate to the first sub category available
         DurationBasedRecommendationSubCategory categoryToConsider = (DurationBasedRecommendationSubCategory) this.category.getRecommendationSubCategories()[0];
         // Loop over categories to set the least category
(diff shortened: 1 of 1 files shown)


src/main/java/com/autotune/analyzer/experiment/ExperimentResultValidation.java
```diff
@@ -64,7 +64,7 @@ public void validate(List<ExperimentResultData> experimentResultDataList, Map<St
 
                     // Check if there is only one K8sObject in resultData
                     if (resultData.getKubernetes_objects().size() > 1) {
-                        errorMsg = errorMsg.concat(String.format("Multiple Kubernetes Objects are detected for result data of experiment : {} .Multiple Kubernetes Objects are unsupported", expName));
+                        errorMsg = errorMsg.concat(String.format("Multiple Kubernetes Objects are detected for result data of experiment: %s .Multiple Kubernetes Objects are unsupported", expName));
```
khansaad, line 67: I think this message can be shortened to something like this:

`Bulk Kubernetes Objects are unsupported!`

It will be consistent with bulk experiment entries response.

khansaad (commented):
This multiple checks can be made concise. But this can be done later when we'll handle all the other validations

src/main/java/com/autotune/analyzer/experiment/ExperimentResultValidation.java
bharathappali: Thanks for the review @khansaad

src/main/java/com/autotune/analyzer/experiment/ExperimentResultValidation.java
```diff
@@ -84,6 +91,57 @@ public void validate(List<ExperimentResultData> experimentResultDataList, Map<St
...
+
+                        // Check if Kubernetes Object Name is matched
+                        String kubeObjNameInKruizeObject = kruizeObject.getKubernetes_objects().get(0).getName();
+                        String kubeObjNameInResultsData = resultData.getKubernetes_objects().get(0).getName();
+
+                        if (!kubeObjNameInKruizeObject.equals(kubeObjNameInResultsData)) {
+                            kubeObjsMisMatch = true;
+                            errorMsg = errorMsg.concat(
+                                    String.format(
+                                            "Kubernetes Object Names MisMatched. Expected Type: %s, Found: %s in Results for experiment: %s \n",
```
dinogun, line 118: Should be `Expected Name`

dinogun (approved):
Lgtm

PR #745, opened 2023-04-14
Create Experiment validation for mandatory params

khansaad:
This PR fixes #559

PR #847, opened 2023-06-05
Fix empty exp name acceptance in update results

bharathappali:
This PR doesn't allow empty experiment name results to be updated.


dinogun (approved):
LGTM. createExperiment will be fixed as part as #745

PR #851, opened 2023-06-06
Exposed kruize service to run on openshift cluster

chandrams:


chandrams:
On scalelab cluster tests are failing as kruize service needs to be exposed. This change is required to test if jenkins pipelines are working on scalelab cluster. 

@dinogun - can you please review this change, thank you.

dinogun (approved):
LGTM

PR #856, opened 2023-06-13
Updated the workflows to include PR number

chandrams:


dinogun (commented):
Looks like the scale test needs to be updated to include these changes? Are there other tests that are using `:test` as well?

chandrams:
@dinogun - Yes, I see it being referenced in stress and fault tolerant test too. I will update it.

dinogun (approved):
LGTM

PR #862, opened 2023-06-19
Renamed the workflow file to keep it consistent

chandrams:
Renamed the workflow file to keep it consistent and to check if the workflow is detected in Actions tab


dinogun (approved):
LGTM

PR #900, opened 2023-08-07
Updated the temporary yaml file name to include cluster name

chandrams:


dinogun (approved):
Thanks for removing the `tmp/` reference. LGTM

PR #903, opened 2023-08-17
Rename duration based to cost

bharathappali:
This PR renames the duration based engine to cost

Also makes the engine to use 95th percentile for CPU and memory to give out the recommendations


src/main/java/com/autotune/analyzer/recommendations/engine/CostRecommendationEngine.java
```diff
@@ -280,11 +279,11 @@ private static RecommendationConfigItem getMemoryRequestRecommendation(Map<Times
                 .collect(Collectors.toList());
 
         // Add a buffer to the current usage max
-        Double memRecUsage = CommonUtils.percentile(HUNDREDTH_PERCENTILE, memUsageList);
+        Double memRecUsage = CommonUtils.percentile(NINETY_FIFTH_PERCENTILE, memUsageList);
         Double memRecUsageBuf = memRecUsage + (memRecUsage * MEM_USAGE_BUFFER_DECIMAL);
 
         // Add a small buffer to the current usage spike max and add it to the current usage max
-        Double memRecSpike = CommonUtils.percentile(HUNDREDTH_PERCENTILE, spikeList);
+        Double memRecSpike = CommonUtils.percentile(NINETY_FIFTH_PERCENTILE, spikeList);
```
dinogun, line 286: Please keep the memory calculations the same as before. Lowering Memory usage might cause the app to crash. Lowering CPU will only cause the app to stall.

dinogun (commented):
1. Please make the appropriate test changes
2. Update all Doc to indicate the "duration based" change to "cost"

src/main/java/com/autotune/analyzer/recommendations/engine/CostRecommendationEngine.java
bharathappali: Reverted the memory value to have `100th percentile`

PR #910, opened 2023-08-22
thread pool count changed from 6 to 10

msvinaykumar:


dinogun (approved):
Approving for now, but we need this to be based on the number of CPUs as seen from inside the container.

PR #913, opened 2023-08-23
latest=false fixed the code

msvinaykumar:
/**
             * Identify timestamps that exhibit the "NOT_ENOUGH_DATA" status
             * for all three categories: short, medium, and long etc. Subsequently,
             * eliminate these timestamps from the k8sObject to prevent the data from being stored in the database.
             * As a result, when using the query parameter "listRecommendation?latest=false,"
             * only a collection of viable recommendations will be displayed
             */


dinogun (approved):
latest changes now looking much better. Approving

PR #914, opened 2023-08-23
Error map for updateResults

msvinaykumar:
will get code with error message in updateResultsAPI

       "errors": {
                "409": "An entry for this record already exists!"
            },

### DIFF
--- a/src/main/java/com/autotune/analyzer/experiment/ExperimentInitiator.java
+++ b/src/main/java/com/autotune/analyzer/experiment/ExperimentInitiator.java
@@ -15,6 +15,7 @@
  *******************************************************************************/
 package com.autotune.analyzer.experiment;
 
+import com.autotune.analyzer.exceptions.KruizeResponse;
 import com.autotune.analyzer.kruizeObject.KruizeObject;
 import com.autotune.analyzer.performanceProfiles.PerformanceProfileInterface.PerfProfileInterface;
 import com.autotune.analyzer.serviceObjects.Converters;
@@ -36,7 +37,10 @@ import org.slf4j.LoggerFactory;
 import javax.servlet.http.HttpServletResponse;
 import java.lang.reflect.Field;
 import java.sql.Timestamp;
-import java.util.*;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
 
 /**
  * Initiates new experiment data validations and push into queue for worker to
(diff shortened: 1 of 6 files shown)


src/main/java/com/autotune/database/service/ExperimentDBService.java
```diff
@@ -231,6 +233,43 @@ public ValidationOutputData addRecommendationToDB(Map<String, KruizeObject> expe
...
+                            } else {
+                                LOGGER.debug("timestamp: {} notEnoughDataCount : {} recommendation size() : {} ", timestampKey, notEnoughDataCnt, recommendationHashMap.size());
+                            }
+                        }
+                    }
+                    emptyTimestamps.forEach((timestamp -> {
+                        cr.getData().remove(timestamp);
+                    }));  // remove timestamp with NOT_ENOUGH_DATA for all three category
+                }
+            }
```
dinogun, line 272: This needs to be removed from here

msvinaykumar: done

dinogun:
The format that we have been using for all the APIs has been as below
```
{
  "message": "Mandatory parameters missing.",
  "httpcode": 400,
  "documentationLink": "",
  "status": "ERROR"
}
```
@msvinaykumar I think we need to follow the same format for each individual record inside the bulk JSON as well. This means that we don't need two different ways to handle error conditions, one type for the entire bulk record and one for individual. Instead we need to have the same format for both.

msvinaykumar:
> The format that we have been using for all the APIs has been as below
> 
> ```
> {
>   "message": "Mandatory parameters missing.",
>   "httpcode": 400,
>   "documentationLink": "",
>   "status": "ERROR"
> }
> ```
> 
> @msvinaykumar I think we need to follow the same format for each individual record inside the bulk JSON as well. This means that we don't need two different ways to handle error conditions, one type for the entire bulk record and one for individual. Instead we need to have the same format for both.

ok let me check

msvinaykumar:
> The format that we have been using for all the APIs has been as below
> 
> ```
> {
>   "message": "Mandatory parameters missing.",
>   "httpcode": 400,
>   "documentationLink": "",
>   "status": "ERROR"
> }
> ```
> 
> @msvinaykumar I think we need to follow the same format for each individual record inside the bulk JSON as well. This means that we don't need two different ways to handle error conditions, one type for the entire bulk record and one for individual. Instead we need to have the same format for both.

Done

src/main/java/com/autotune/analyzer/performanceProfiles/PerformanceProfileInterface/ResourceOptimizationOpenshiftImpl.java
```diff
@@ -177,6 +174,44 @@ public void generateRecommendation(KruizeObject kruizeObject, List<ExperimentRes
...
+                                if (notEnoughDataCnt == recommendationHashMap.size()) {
+                                    emptyTimestamps.add(timestampKey); // found timestamp with NOT_ENOUGH_DATA for all three category
+                                } else {
+                                    LOGGER.debug("timestamp: {} notEnoughDataCount : {} recommendation size() : {} ", timestampKey, notEnoughDataCnt, recommendationHashMap.size());
+                                }
+                            }
+                        }
+                        emptyTimestamps.forEach((timestamp -> {
+                            cr.getData().remove(timestamp);
+                        }));  // remove timestamp with NOT_ENOUGH_DATA for all three category
```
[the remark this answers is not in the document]
msvinaykumar: sorry Dino I did not get . Do you want me to move this code somewhere else ?

dinogun (approved):
LGTM

Issue #928, opened 2023-09-04
[DB Scalability] Java changes corresponding to the Cluster_Name and Indexes


khansaad:
Java related changes needed for the DB Scalability updates:

1. Add cluster_name in all SQLs
2. Create indexes in the results and recommendations table

rbadagandi1:
This is completed as part of V0.0.19 release

PR #931, opened 2023-09-05
Add indices for clusterName and interval endtime in results table

khansaad:
This PR adds the indices for clusterName and interval endtime in results table as part of the epic https://github.com/kruize/autotune/issues/928


msvinaykumar (approved):
LGTM

PR #948, opened 2023-09-15
Update Partition Creation Strategy

khansaad:
This PR updates the partition creation strategy.

- Changed logic to create partition for the rest of the month from the day on which the results are inserted.


dinogun:
@khansaad does this take care of the scenario where the results come out of order? Eg. If I get results for say 10-Sep first and then 09-Sep

msvinaykumar (approved):

PR #950, opened 2023-09-15
DB partition changes

msvinaykumar:


src/main/java/com/autotune/analyzer/services/UpdateResults.java
```diff
@@ -59,10 +59,14 @@ public class UpdateResults extends HttpServlet {
...
     private static final Logger LOGGER = LoggerFactory.getLogger(UpdateResults.class);
     public static ConcurrentHashMap<String, PerformanceProfile> performanceProfilesMap = new ConcurrentHashMap<>();
+    public static ConcurrentHashMap<String, KruizeObject> mainKruizeExperimentMAP;
 
     @Override
     public void init(ServletConfig config) throws ServletException {
         super.init(config);
+        mainKruizeExperimentMAP = (ConcurrentHashMap<String, KruizeObject>) config.getServletContext().getAttribute(AnalyzerConstants.EXPERIMENT_MAP);
+        if (mainKruizeExperimentMAP == null)
+            mainKruizeExperimentMAP = new ConcurrentHashMap<>();
```
dinogun, line 69: What is the scope of this object?  Why is this not part of `doPost`

src/main/java/com/autotune/utils/KruizeConstants.java
```diff
@@ -189,6 +189,7 @@ public static final class JSONKeys {
         public static final String INTERVAL_START_TIME = "interval_start_time";
```
dinogun, line 189: Is this file really needed for this PR? Not really important but this seems to be part of the perf profile changes

dinogun (commented):
@msvinaykumar to ensure that the ROS team does not have to make any changes in the 0.0.19_rm release, I'd suggest that the `cluster_name` has to be non-null in update_results NOT be enforced in this PR. We can enforce it in 0.0.20 along with both createExperiment and updateResults at the same time

dinogun (commented):
Looks like the migrations folder is missing

src/main/java/com/autotune/analyzer/services/UpdateResults.java
msvinaykumar: removed

dinogun (approved):
LGTM

PR #956, opened 2023-09-18
Adds threshold for limit rows

bharathappali:
This PR adds threshold for `limit rows` to make sure duration in mins is covered for all terms.


dinogun (approved):
LGTM

PR #966, opened 2023-09-26
fixed when cluster_name is null

msvinaykumar:
Correction pertaining to 'cluster_name' whether it is null or not null.


dinogun (approved):
LGTM

PR #967, opened 2023-09-26
Script to gather Kruize metrics

kusumachalasani:
Python script to gather kruize metrics


tests/scripts/kruize_metrics.py
```diff
@@ -0,0 +1,357 @@
...
+        elif opt == '-r':
+            resultsfile = arg
+            
+    if '-t' not in sys.argv:
+        time_duration = "60m"
+    if '-q' not in sys.argv:
+        queries_type = "increase"
+    if '-d' not in sys.argv:
+        duration = None
+    if '-p' not in sys.argv:
```
dinogun, line 270: Make this as the default as a casual user just wants to know what the data looks like

tests/scripts/kruize_metrics.py
```diff
@@ -0,0 +1,357 @@
...
+    global time_duration 
+    global queries_map
+    global getOneDataPoint
+    global resultsfile
+    global namespace
+
+    try:
+        opts, args = getopt.getopt(argv,"h:c:s:d:t:q:p:r:")
+    except getopt.GetoptError:
+        print("kruize_metrics.py -c <cluster type> -s <server>")
```
dinogun, line 243: If you dont provide any parameters, the script should print the help message

tests/scripts/kruize_metrics.py
```diff
@@ -0,0 +1,357 @@
...
+    if '-t' not in sys.argv:
+        time_duration = "60m"
+    if '-q' not in sys.argv:
+        queries_type = "increase"
+    if '-d' not in sys.argv:
+        duration = None
+    if '-p' not in sys.argv:
+        getOneDataPoint = "false"
+    if '-r' not in sys.argv:
+        resultsfile = "kruizemetrics.csv"
```
dinogun, line 273: Both at the start and completion of the script, the output file should be mentioned

tests/scripts/kruize_metrics.py
dinogun: It is not clear what is the server here for a first time user, please elaborate in the help message

dinogun: Since cluster type is fixed, print the options

dinogun: If it is unable to access the server port, it should catch the error and print a helpful message as this would be the most common error

tests/scripts/kruize_metrics.py
kusumachalasani: Done.

tests/scripts/kruize_metrics.py
kusumachalasani: Done.

dinogun:
@kusumachalasani The `server` parameter needs to be explained more ... it is still not clear

tests/scripts/kruize_metrics.py
```diff
@@ -0,0 +1,359 @@
+"""
```
dinogun, line 1: Please move this to the ${ROOT_DIR}/scripts folder and not tests

dinogun (approved):
LGTM

PR #968, opened 2023-09-27
optimal sql

msvinaykumar:
optimal query
        select *  from kruize_results k1_0 where k1_0.experiment_name=? and k1_0.interval_end_time<=(updateRecommendation API given interval_end-time) and interval_end_time>= START_TIME

here START_TIME = interval_end_time - (LONG_TERM_DURATION_DAYS + THRESHOLD 2 days)

### DIFF
--- a/src/main/java/com/autotune/analyzer/performanceProfiles/PerformanceProfileInterface/ResourceOptimizationOpenshiftImpl.java
+++ b/src/main/java/com/autotune/analyzer/performanceProfiles/PerformanceProfileInterface/ResourceOptimizationOpenshiftImpl.java
@@ -93,13 +93,13 @@ public class ResourceOptimizationOpenshiftImpl extends PerfProfileImpl {
         cal.add(Calendar.DAY_OF_MONTH, -(KruizeConstants.RecommendationEngineConstants.DurationBasedEngine.DurationAmount.LONG_TERM_DURATION_DAYS +
                 KruizeConstants.RecommendationEngineConstants.DurationBasedEngine.DurationAmount.LONG_TERM_DURATION_DAYS_THRESHOLD));
         // Get the new Timestamp after subtracting 10 days
-        interval_start_time = new Timestamp(cal.getTimeInMillis());
+        Timestamp calculated_start_time = new Timestamp(cal.getTimeInMillis());
         Map<String, KruizeObject> mainKruizeExperimentMap = new HashMap<>();
         String experiment_name = kruizeObject.getExperimentName();
         mainKruizeExperimentMap.put(experiment_name, kruizeObject);
         new ExperimentDBService().loadResultsFromDBByName(mainKruizeExperimentMap,
                 experiment_name,
-                interval_start_time,interval_end_time);
+                calculated_start_time,interval_end_time);
         //TODO: Will be updated once algo is completed
         for (ExperimentResultData experimentResultData : experimentResultDataList) {
             if (null != kruizeObject && null != experimentResultData) {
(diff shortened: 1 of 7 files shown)


src/main/java/com/autotune/analyzer/performanceProfiles/PerformanceProfileInterface/ResourceOptimizationOpenshiftImpl.java
```diff
@@ -83,36 +80,26 @@ public List<KruizeRecommendationEngine> getEngines() {
...
+        // Get the new Timestamp after subtracting 10 days
+        interval_start_time = new Timestamp(cal.getTimeInMillis());
         Map<String, KruizeObject> mainKruizeExperimentMap = new HashMap<>();
+        String experiment_name = kruizeObject.getExperimentName();
         mainKruizeExperimentMap.put(experiment_name, kruizeObject);
         new ExperimentDBService().loadResultsFromDBByName(mainKruizeExperimentMap,
                 experiment_name,
-                interval_end_time,
-                practicalLimitRows);
+                interval_start_time,interval_end_time);
```
dinogun, line 102: interval_start_time already has a different meaning, so this needs to be called something else

src/main/java/com/autotune/analyzer/performanceProfiles/PerformanceProfileInterface/ResourceOptimizationOpenshiftImpl.java
```diff
@@ -83,36 +80,26 @@ public List<KruizeRecommendationEngine> getEngines() {
...
+        // Convert the Timestamp to a Calendar instance in UTC time zone
+        Calendar cal = Calendar.getInstance(TimeZone.getTimeZone("UTC"));
+        cal.setTimeInMillis(interval_end_time.getTime());
+        /*
+         * interval_start_time Subtract (LONG_TERM_DURATION_DAYS +  THRESHOLD days)
+         * Incorporate a buffer period of "threshold days" to account for potential remote cluster downtime.
+         * This adjustment aims to align the cumulative hours' duration with LONG_TERM_DURATION_DAYS.
+         */
+        cal.add(Calendar.DAY_OF_MONTH, -(KruizeConstants.RecommendationEngineConstants.DurationBasedEngine.DurationAmount.LONG_TERM_DURATION_DAYS +
+                KruizeConstants.RecommendationEngineConstants.DurationBasedEngine.DurationAmount.LONG_TERM_DURATION_DAYS_THRESHOLD));
```
dinogun, line 94: I think rather than adding a threshold and hoping it will work for all cases, I think this needs to be converted to a loop. After each iteration we can check if we have the required number of hours, if not we try again with the same quantum again by going further back in time. This can stop after we have reached a limit or there are no more rows, in which case, we return the "NOT_ENOUGH_DATA" error

src/main/java/com/autotune/database/dao/ExperimentDAO.java
```diff
@@ -50,7 +50,7 @@ public interface ExperimentDAO {
 
     // Load all results for a particular experimentName
 
-    List<KruizeResultsEntry> loadResultsByExperimentName(String experimentName, String cluster_name, Timestamp interval_start_time, Integer limitRows) throws Exception;
+    List<KruizeResultsEntry> loadResultsByExperimentName(String experimentName, String cluster_name, Timestamp interval_start_time,  Timestamp interval_end_time) throws Exception;
```
dinogun, line 53: Can you please do the same nomenclature for interval_start_time everywhere?

dinogun (approved):
LGTM

PR #969, opened 2023-09-28
Fix term threshold

bharathappali:
n order to enhance the recommendation generation process, this PR builds upon the existing logic outlined in [PR #968](https://github.com/kruize/autotune/pull/968). The primary objective is to introduce the concept of maximum data thresholds for each term, where the threshold is defined as the TERM_DURATION plus an additional 6 hours of data.

To achieve this, the proposed modification focuses on optimizing the monitoringStartTime function. Instead of simply retrieving a list of timestamps until the monitoring start time (Calculated based on sum of duration in mins value), the function has been adjusted to provide a value that aligns with the defined data thresholds. This adjustment ensures that the recommendation generation process remains consistent while accommodating the new data threshold requirements.

PS: This PR is built on top of #968 so it should be merged before merging this PR.

### DIFF
--- a/src/main/java/com/autotune/analyzer/recommendations/RecommendationConstants.java
+++ b/src/main/java/com/autotune/analyzer/recommendations/RecommendationConstants.java
@@ -654,6 +654,10 @@ public class RecommendationConstants {
         public static final Double MEM_SPIKE_BUFFER_DECIMAL = 0.05;
         public static final Double DEFAULT_CPU_THRESHOLD = 0.1;
         public static final Double DEFAULT_MEMORY_THRESHOLD = 0.1;
+
+        public static final int THRESHOLD_HRS_SHORT_TERM = 6;
+        public static final int THRESHOLD_HRS_MEDIUM_TERM = 6;
+        public static final int THRESHOLD_HRS_LONG_TERM = 6;
     }
 
     public static final class RecommendationNotificationMsgConstant {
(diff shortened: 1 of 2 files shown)


src/main/java/com/autotune/analyzer/recommendations/RecommendationConstants.java
```diff
@@ -654,6 +654,10 @@ public static final class RecommendationValueConstants {
         public static final Double MEM_SPIKE_BUFFER_DECIMAL = 0.05;
         public static final Double DEFAULT_CPU_THRESHOLD = 0.1;
         public static final Double DEFAULT_MEMORY_THRESHOLD = 0.1;
+
+        public static final int THRESHOLD_HRS_SHORT_TERM = 6;
```
msvinaykumar, line 658: is it worth setting this threshold from app config ? so that code change is not required.

msvinaykumar (approved):
I've examined this code within my local environment, and it appears to be functioning correctly. We can schedule a call to confirm the setup and test the cases as we proceed.

src/main/java/com/autotune/analyzer/recommendations/RecommendationConstants.java
bharathappali: Yes that's a great idea, we can have an app config entry for this with which we can make user to tune it and make these entries as default values.

dinogun (approved):
LGTM

PR #970, opened 2023-09-28
added execute permission

msvinaykumar:


manifests/crc/default-db-included-installation/minikube/kruize-crc-minikube.yaml
```diff
@@ -199,6 +199,7 @@ apiVersion: batch/v1
 kind: CronJob
 metadata:
   name: create-partition-cronjob
+  namespace: openshift-tuning
```
khansaad, line 202: I think the namespace in minikube would be different.

dinogun (approved):
LGTM

PR #973, opened 2023-09-29
Add docs for the term threshold changes

bharathappali:
This PR formats the JSON to be in expandable sections and add the documentation about threshold incorporated in considering the data analysed for a particular term to provide recommendations


dinogun (commented):
@bharathappali can you add a TOC at the top so it is easy for users to find what they want?

dinogun (approved):
Looking good now

PR #976, opened 2023-10-04
Rm merge mvp

msvinaykumar:


manifests/crc/default-db-included-installation/openshift/kruize-crc-openshift.yaml
```diff
@@ -248,6 +248,47 @@ spec:
...
+  namespace: openshift-tuning
+spec:
+  schedule: "0 0 25 * *" # Run on 25th of every month at midnight
+  jobTemplate:
+    spec:
+      template:
+        spec:
+          containers:
+            - name: kruizecronjob
+              image: kruize/autotune_operator:0.0.19_rm
```
dinogun, line 264: This needs to be 0.0.19.2_rm as well (I guess this needs to be fixed in rm as well)

dinogun (approved):
LGTM

PR #979, opened 2023-10-11
Included a workflow to run functional tests on commenting on a PR

chandrams:


dinogun (approved):
Approving as this is the same as that in mvp_demo

PR #990, opened 2023-10-14
HV Error fix

msvinaykumar:
Trying to fix Hv000149
By refering to

https://stackoverflow.com/questions/65720131/an-exception-occurred-during-message-interpolation-spring-validation


dinogun (approved):
Good find!

PR #1009, opened 2023-10-27
Included tests to update results to validate behavior when metrics section is missing

chandrams:
Included tests to update results to validate behavior when metrics section is missing


chandrams:
Have raised this directly from kruize repo, will close this and raise from my fork

PR #1010, opened 2023-10-27
Included tests for validating the behavior when metrics section is missing

chandrams:
Included tests for validating the behavior when metrics section is missing


dinogun (approved):
LGTM

PR #1016, opened 2023-11-02
Updated error checking in pod log and pod log names before and after restart

chandrams:
Updated error checking in pod log and pod log names before and after restart


dinogun (approved):
LGTM

Issue #1022, opened 2023-11-03
Variation issue in the list recommendations


bharathappali:
@khansaad has pointed out that there was an issue with variation as the variation values are not matching the difference between current and generated values

```
test_list_recommendations.py:1187: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

current_config = {'limits': {'cpu': {'amount': 1.5, 'format': 'cores'}, 'memory': {'amount': 100.0, 'format': 'MiB'}}, 'requests': {'memory': {'amount': 50.21, 'format': 'MiB'}}}
recommended_config = {'limits': {'cpu': {'amount': 1.03, 'format': 'cores'}, 'memory': {'amount': 238.2, 'format': 'MiB'}}, 'requests': {'cpu': {'amount': 1.03, 'format': 'cores'}, 'memory': {'amount': 238.2, 'format': 'MiB'}}}
variation_config = {'limits': {'cpu': {'amount': 1.03, 'format': 'cores'}, 'memory': {'amount': 138.2, 'format': 'MiB'}}, 'requests': {'cpu': {'amount': 1.03, 'format': 'cores'}, 'memory': {'amount': 187.98999999999998, 'format': 'MiB'}}}
```

bharathappali:
Fixed in #1023

PR #1023, opened 2023-11-03
Make CPU Limits to point correctly, earlier it's pointing to cpu request

bharathappali:
As mentioned in issue #1022 the variation is not matching the difference between current and recommended value. The issue is due to the incorrect mapping of cpu limits to cpu requests.

@dinogun Can I please have your review?

### DIFF
--- a/src/main/java/com/autotune/analyzer/recommendations/engine/CostRecommendationEngine.java
+++ b/src/main/java/com/autotune/analyzer/recommendations/engine/CostRecommendationEngine.java
@@ -993,7 +993,7 @@ public class CostRecommendationEngine implements KruizeRecommendationEngine {
             HashMap<String, RecommendationConfigItem> internalMapToPopulate = new HashMap<String, RecommendationConfigItem>();
             // Add current values
             internalMapToPopulate.put(RecommendationConstants.RecommendationEngine.InternalConstants.CURRENT_CPU_REQUEST, currentCPURequest);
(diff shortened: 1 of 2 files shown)


bharathappali:
Thanks @khansaad for pointing the issue

dinogun:
Good catch indeed @khansaad!!

dinogun (commented):
Do we need to do this calculation in both the engines? Can this be done outside in common code

bharathappali:
@dinogun It is jus calculating a difference so I have added it in the engines, Can we extracted and common out.

bharathappali:
> Do we need to do this calculation in both the engines?

Yes, calculations are done in both the engines. 

> Can this be done outside in common code

Yes can be done, but after a detailed inspection I felt it would be better to have these calculations done in the respective engines as these calculations are tightly coupled to populating the notifications and also lot many objects need to be sent to the common code to take decisions and calculate the differences, Moreover these engines using the data for representational purpose as of now, but in future if engine needs to take some actions or need to tweak certain things in the data it would be hard to do it in common code as we need to pass the engine name and implement engine specific behaviour over there which might be not correct considering the abstractions built around engines to do the engine specific work in their scope. Please lemme know your thoughts on this.

dinogun:
ok, we can schedule the refactoring for a later time then

dinogun (approved):
LGTM

PR #1024, opened 2023-11-05
Update kruize metrics document

kusumachalasani:
Update document about the script to collect kruize metrics and the output.


dinogun (approved):
LGTM

PR #1037, opened 2023-11-10
Updated a couple of missing mandatory fields for create exp json

chandrams:


dinogun:
@msvinaykumar can you please review. Not sure if these createExperiment fields should be mandatory

msvinaykumar:
Seems we need design discussion , Right now it is not manadatory

chandrams:
Closing this as per the comments above.

Issue #1036, opened 2023-11-09
Add support for different Units for the metric values


khansaad:
Currently Kruize supports only `MiB` unit for the metric values. We need to add support for different memory units like `Bytes, KiB, GiB`

rbadagandi1:
This issue is fixed as part of PR 1038.

PR #1038, opened 2023-11-10
Added support for additional units for the metric values

shekhar316:
This PR solves Issue https://github.com/kruize/autotune/issues/1036

- Currently Kruize supports only MiB unit for the metric values. We added support for different memory units like Bytes, KiB, MiB, GiB, TiB, PiB, EiB

### DIFF
--- a/src/main/java/com/autotune/utils/KruizeSupportedTypes.java
+++ b/src/main/java/com/autotune/utils/KruizeSupportedTypes.java
@@ -78,7 +78,7 @@ public class KruizeSupportedTypes
 			new HashSet<>(Arrays.asList("deployment", "pod", "container"));
 	
   public static final Set<String> SUPPORTED_FORMATS =
-			new HashSet<>(Arrays.asList("cores", "MiB"));
+			new HashSet<>(Arrays.asList("cores", "m", "Bytes", "bytes", "KiB", "MiB", "GiB", "TiB", "PiB", "EiB", "Ki", "Mi", "Gi", "Ti", "Pi", "Ei", "kB", "KB", "MB", "GB", "TB", "PB", "EB", "K", "k", "M", "G", "T", "P", "E"));
 
 	public static final Set<String> QUERY_PARAMS_SUPPORTED = new HashSet<>(Arrays.asList(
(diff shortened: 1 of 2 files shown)


src/main/java/com/autotune/utils/KruizeSupportedTypes.java
```diff
@@ -78,7 +78,7 @@ private KruizeSupportedTypes() { }
 			new HashSet<>(Arrays.asList("deployment", "pod", "container"));
 	
   public static final Set<String> SUPPORTED_FORMATS =
-			new HashSet<>(Arrays.asList("cores", "MiB"));
+			new HashSet<>(Arrays.asList("cores", "Bytes", "KiB", "MiB", "GiB"));
```
dinogun, line 81: Please take a look at the supported kubernetes resource units as per the official documentation
https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/

We need to support all of the above

shekhar316: Hi @dinogun , 

As per the documentation, I have added the all missing formats and modified the PR. Please let me know your comments. 

Thanks

shekhar316: Hi @dinogun, 

Updated the code for all supported formats. Please review again.

dinogun:
@shekhar316 we could have `bytes` as well (All lower case)

dinogun:
@shekhar316 Please dont use force push for your commits, it's not a good practice as it is difficult to look at what changed between commits.

dinogun (approved):
LGTM

dinogun:
@chandrams what tests need to be added / changed to now support the `Bytes` format?

shekhar316:
> @shekhar316 Please dont use force push for your commits, it's not a good practice as it is difficult to look at what changed between commits.

Sure, will keep in mind from next time ...

chandrams:
@dinogun - The results json used in the tests at present have format as MiB with some dummy metrics values, we can change the memory format with the same value to the supported formats and test it. 

Recommendation infra uses real data metrics we can update the scripts there as well to test the actual recommendations with these supported formats.

PR #1041, opened 2023-11-16
Remove check for threshold in returning monitoring start time

### LINKED ISSUE #1040: No recommendation object where parent notification object says `Duration Based Recommendations Available`  but actually there is no recommendation available in the inner json object
incorrect recommendation object from kruize where parent notification object says `Duration Based Recommendations Available`  but actually there is no recommendation available in the inner json object.

https://issues.redhat.com/browse/RHIROS-1427

bharathappali:
Fixes #1040


dinogun (approved):
LGTM

PR #1051, opened 2023-11-21
Fix UpdateResultsAPI duplicate record issue

khansaad:
This PR fixes the duplicate record mismatch issue after hitting the updateResults API. We were seeing a difference in the number of records getting saved in the DB.

PR #1058, opened 2023-11-23
Fix the duplicate record issue in update results API

khansaad:
This PR fixes the duplicate record mismatch issue after hitting the updateResults API. We were seeing a difference in the number of records getting saved in the DB.

Note: Created this separate PR against the 0.0.19 branch. PR #1051 was raised against master.


dinogun (approved):
Same as PR #1051 but against `0.0.19` branch

Issue #1065, opened 2023-11-27
After DB size increase create experiment fails


msvinaykumar:
Stage  requested to increase DB size using MR https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/89197
after db size increased 
Create experiment failed 

`2023-11-2714:13:52.744 ERROR [qtp88397182-37][ExperimentDAOImpl.java(67)]-Not able to save experiment due to Converting `org.hibernate.exception.SQLGrammarException` to JPA `PersistenceException` : could not execute statement
2023-11-2714:13:52.744 DEBUG [qtp88397182-37][CreateExperiment.java(208)]-[{"version":"1.0","experiment_name":"11789772|514277|test_cost_iqe_ros_ocp_27_11_2023|project-ros-B2|daemonset|pod-ros-B21","performance_profile":"resource-optimization-openshift","mode":"monitor","target_cluster":"remote","kubernetes_objects":[{"type":"daemonset","name":"pod-ros-B21","namespace":"project-ros-B2","containers":[{"container_image_name":"space-heart","container_name":"pod-ros-B21","recommendations":{}}]}],"trial_settings":{"measurement_duration":"15min"},"recommendation_settings":{"threshold":"0.1"}}]
2023-11-2714:13:52.746 ERROR [qtp88397182-37][KruizeErrorHandler.java(66)]-{"message":"Converting `org.hibernate.exception.SQLGrammarException` to JPA `PersistenceException` : could not execute statement","httpcode":400,"documentationLink":"","status":"ERROR"}
2023-11-2714:20:17.027 DEBUG [qtp88397182-32][PerformanceProfileUtil.java(140)]-Added PerformanceProfile: resource-optimization-openshift
2023-11-2714:20:17.027 DEBUG [qtp88397182-32][ExperimentInterfaceImpl.java(47)]-Added Experiment name : 11789772|514278|test_cost_iqe_ros_ocp_27_11_2023|project-ros-A1|deploymentconfig|pod-ros-A11 into main map.
2023-11-2714:20:17.033 ERROR [qtp88397182-32][SqlExceptionHelper.java(138)]-ERROR: column "datasource" is of type jsonb but expression is of type character varying
Hint: You will need to rewrite or cast the expression.
Position: 189
[...]

PR #1071, opened 2023-12-04
JSONB, is not being set to its default value

msvinaykumar:

created issue for same

https://github.com/kruize/autotune/issues/1065


msvinaykumar:
@chandrams this issue not able to reproduce locally  as it requires DB size increase to existing DB without clean up command.

migrations/kruize_experiments_ddl.sql
```diff
@@ -1,7 +1,7 @@
...
-create table IF NOT EXISTS kruize_results (interval_start_time timestamp(6) not null, interval_end_time timestamp(6) not null, experiment_name varchar(255) not null, cluster_name varchar(255) , duration_minutes float(53) not null, extended_data jsonb, meta_data jsonb, version varchar(255), primary key (experiment_name, interval_end_time, interval_start_time)) PARTITION BY RANGE (interval_end_time);
+create table IF NOT EXISTS kruize_recommendations (interval_end_time timestamp(6) not null, experiment_name varchar(255) not null, cluster_name varchar(255), extended_data jsonb , version varchar(255), primary key (experiment_name, interval_end_time)) PARTITION BY RANGE (interval_end_time);
+create table IF NOT EXISTS kruize_results (interval_start_time timestamp(6) not null, interval_end_time timestamp(6) not null, experiment_name varchar(255) not null, cluster_name varchar(255) , duration_minutes float(53) not null, extended_data jsonb, meta_data jsonb DEFAULT NULL, version varchar(255), primary key (experiment_name, interval_end_time, interval_start_time)) PARTITION BY RANGE (interval_end_time);
```
chandrams, line 4: Do we need to set DEFAULT NULL or extended_data jsonb as well?

msvinaykumar: no extended_data cannot be set  DEFAULT NULL, It should have values always.

chandrams (approved):
LGTM

src/main/java/com/autotune/database/helper/DBHelpers.java
```diff
@@ -312,6 +314,7 @@ public static KruizeResultsEntry convertExperimentResultToExperimentResultsTable
                             Double.valueOf((experimentResultData.getIntervalEndTime().getTime() -
                                     experimentResultData.getIntervalStartTime().getTime()) / (60 * 1000))
                     );
+                    kruizeResultsEntry.setMeta_data(null);
```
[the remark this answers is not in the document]
msvinaykumar: Yes, however, the 'datasource' element has not been implemented at the code level; it was only conceptualized during the design phase. To verify the existence of the datasource, we would need to declare that variable in the API object and handle JSON parsing accordingly. All these modifications can be implemented once the datasource is active. Currently, we have chosen to minimize changes to focus on bug fixes.

msvinaykumar:
merged into MVP 
https://github.com/kruize/autotune/pull/1085

PR #1075, opened 2023-12-12
Add confidence level for each recommendation terms

### LINKED ISSUE #1059: Update Term Design and add Confidence Level
The current design [v0.0.20] related to terms has a couple of issues
1. Term durations are not strictly enforced, we can go back in time (upto a threshold level) to satisfy requirements related to a term duration. This means that the user is not clear what duration was considered to arrive at a recommendation
2. If enough data is not available that satisfies the entire term duration, then no recommendation is generated. eg, for short term, we need a full 96 data points with a measurement duration of 15 mins. This means that for any workload that is short lived, say 3 hours, we cannot generate any recommendations

To prevent this, we need to update the term design as follows
1. Fix term durations as follows (Assuming a measurement duration of 15 mins: short term = 24 hours (Max: 96 data points), medium term = 7 days (Max: 672  data points), long term = 15 days (Max: 1440 data points).
2. Remove buffer thresholds. This means that we do not go further back in time, beyond the prescribed term durations above, to get data for a given duration
3. Add a minimum data threshold. This represents the minimum amount of data needed for generating a recommendation associated with a given duration term. We can start with the following defaults. Minimum data points for: Short term: 30 mins (Min: 2 data points), Medium term: 2 days (Min: 192 data points), Long Term: 8 days (Min: 768 data points)
4. The design should allow for the minimum thresholds to be configurable for each individual experiment.
5. Add a confidence level associated with a recommendation. The confidence level is a value between 0 and 1 where 1 represents the highest confidence in the recommendation. The initial implementation will be based entirely on the number of data points used to generate the recommendation. It can be `confidence_level = actual_data_points / term_max_data_points`
6.

khansaad:
This PR adds the confidence level feature of #1059 . PR  #1081 takes care of the term related changes.

- Made changes to calculate the confidence level.

### DIFF
--- a/design/MonitoringModeAPI.md
+++ b/design/MonitoringModeAPI.md
@@ -60,6 +60,13 @@ Below are the default terms used in resource analysis, along with their respecti
 
 **Threshold**: The "threshold" is an additional buffer period. It encompasses the term duration and extends an extra time duration. This buffer accommodates any potential data gaps or misses within the term window, ensuring a comprehensive analysis.
 
+### Confidence Level
+Confidence level is associated with a recommendation. It is a value between `0` and `1` where `1` represents the highest confidence in the recommendation. Our initial implementation is based entirely on the number of data points used to generate the recommendation. 
+It is calculated as: 
+```
+confidence_level = actual_data_points / term_max_data_points
+```
+**Note**: This is a work-in-progress feature. We'll update it iteratively based on the design discussions.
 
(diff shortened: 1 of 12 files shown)


khansaad:
> @khansaad - Can you update the docs and list the test scenarios for me to review. Are you raising a separate PR with the test changes?

Updated the doc file and added a minor test change. Kindly review once.

PR #1077, opened 2023-12-13
JSONB, is not being set to its default value

msvinaykumar:

created issue for same

https://github.com/kruize/autotune/issues/1065


msvinaykumar:
mered into MVP https://github.com/kruize/autotune/pull/1085

PR #1085, opened 2024-01-19
Handle null after space increase

msvinaykumar:
Merging from RM to MVP

https://github.com/kruize/autotune/pull/1077
https://github.com/kruize/autotune/pull/1071


dinogun:
@msvinaykumar can you please indicate how this has been tested, thanks

msvinaykumar:
> @msvinaykumar can you please indicate how this has been tested, thanks

this issue not able to reproduce locally as it requires DB size increase to existing DB without clean up command.

Normal testcases should pass.

dinogun:
@chandrams can you review this

chandrams:
@dinogun - I have tested this PR with the kruize release 0.0.19.5 with 30 exps / 15 days and updated Recommendations, with the same postgres DB restarted only kruize with the build containing this PR 1085, posted a day's results for the same 30 exps and invoked updateRecommendations. Uploading results and recommendations generation worked as expected.

chandrams (approved):
LGTM

PR #1118, opened 2024-03-06
fetch token for prometheus access and add CRB

msvinaykumar:


src/main/java/com/autotune/utils/GenericRestApiClient.java
```diff
@@ -59,14 +62,29 @@ public class GenericRestApiClient {
 
     /**
      * Initializes a new instance just by passing baseURL which does not need any authentication.
+     *
      * @param baseURL
      */
     public GenericRestApiClient(String baseURL) {
         this.baseURL = baseURL;
+        String tokenFilePath = "/var/run/secrets/kubernetes.io/serviceaccount/token";
```
msvinaykumar, line 70: Need to check if this token do rotation after expiry

dinogun: A few comments on this approach
1. We need to move this out of utils and have a separate package
2. Generic version can take a token and not assume any specific token file path
3. Openshift version can extend the generic version and specify the token file path

dinogun (approved):
Hackathon merge!

PR #1122, opened 2024-03-13
Updating prometheus-kube version to fix the issue with deprecated resources

ncau:
Changed the kube-prometheus clone to the latest version and changed applys to creates to avoid issues with customResourceDefintition length. Also had to change the applys to creates in order to avoid issues with custom resource definition lengths.

There might still be issues with the github runners so checking this here.

### DIFF
--- a/scripts/prometheus_on_minikube.sh
+++ b/scripts/prometheus_on_minikube.sh
@@ -22,9 +22,6 @@ non_interactive=0
 # Call setup by default (and not terminate)
 setup=1
 
-# Prometheus release default
-tag="v0.8.0"
-
 function install_prometheus() {
 	echo
 	echo "Info: Checking pre requisites for prometheus..."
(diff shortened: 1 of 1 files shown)


ncau:
Seems were running 1.19 of kubernetes from what I found after 1.25 the relevant packages are fully deprecated (not 100% sure from what version on minikube they exceeded 1.25 but 1.30.1 minikube is running 1.26.3). Is this a case of we individually should run older minikube/kubernetes builds or is it something that needs to be bumped in the pipeline @dinogun?

ncau:
Alternatively assuming we dont want to update autotune to be compatible post 1.25, we can do a check for the kubernetes version via kubectl.  Then force the user to try with the final version of minikube prior to 1.26 (1.25.2) by spawning the minikube instance with --version 1.25.2 in the common helper(I have already written this if preferred). There is currently a check done but only in the hpo_demo_setup and it extracts the version from minikube directly instead of via kubectl which means that even if you run the correct version of kubernetes via minikube with a --version command it will fail.

PR #1124, opened 2024-03-19
Troubleshooting EE Perfprofile error and avoid Pod restart solution.

msvinaykumar:

PR #1141, opened 2024-03-27
Fix Postgres Connection Count

khansaad:
This PR fixes the issue where we were seeing an increase in the active postgres connections after PR #1124 merge.


dinogun (approved):
LGTM

PR #1155, opened 2024-04-10
Merge recommendation generation code

khansaad:
This PR aims to remove the redundancies in the recommendation code present in `kruize-local` and `mvp_demo` branch.

- Merged `kruize-local` code to make the recommendation generation common and removed the `GenerateRecommendation` API.
- `UpdateRecommendation` API will be used for both remote and local monitoring usecases.
- Reverted the `datasource` datatype change in the `kruize_experiments` table back to `jsonb`
-  Updated the corresponding API object and related classes to handle the behavior of datasource object when it's sent during createExperiment stage.


khansaad:
Raised #1166  to include above changes. This is no longer required

Issue #841, opened 2023-06-02
Include tests for validating the notification codes into the kruize functional test suite


chandrams:
Include tests for validating the [notification codes](https://github.com/kruize/autotune/blob/mvp_demo/design/NotificationCodes.md) into the kruize functional test suite

bharathappali:
Tests added for 
```
 INVALID_AMOUNT_IN_CPU_SECTION_CODE = "223002"
 INVALID_FORMAT_IN_CPU_SECTION_CODE = "223004"
 INVALID_AMOUNT_IN_MEMORY_SECTION_CODE = "224002"
 INVALID_FORMAT_IN_MEMORY_SECTION_CODE = "224004"
```

Need to create scenarios and test cases for missing use cases

PR #1156, opened 2024-04-11
Updated to trigger a notification if CPU recommendations are unavailable at profile level.

msvinaykumar:
Updated to trigger a notification if CPU recommendations are unavailable at profile level.

following notifications are missed

 "CPU Usage is less than a millicore, No CPU Recommendations can be generated"
 "CPU usage is zero, No CPU Recommendations can be generated"


bharathappali (approved):
LGTM

But I feel we need to have unified notification structure in future.

I see `engineNotifications` object is created at same level and notifications are populated to that and at the end they are being added to `recommendationModel`  if we have `notifications` object passed we should make use of the same object instead of creating a new one.

chandrams:
@msvinaykumar - Can you please update the function test suite with tests to ensure such regressions are not missed out. You can refer this [issue](https://github.com/kruize/autotune/issues/841) to check for tests added for which notifications. Thank you.

msvinaykumar:
> @msvinaykumar - Can you please update the function test suite with tests to ensure such regressions are not missed out. You can refer this [issue](https://github.com/kruize/autotune/issues/841) to check for tests added for which notifications. Thank you.

@chandrams  was there any previously written test cases check for CPU usage being zero or less than a millicore. Or Do I need to include new test cases?

chandrams:
@msvinaykumar - Tests for these notifications aren't present in the testsuite, new tests are required. You can check the issue above to see for which notification codes tests were added earlier.

src/main/java/com/autotune/analyzer/recommendations/engine/RecommendationEngine.java
```diff
@@ -1228,6 +1228,11 @@ private boolean populateRecommendation(Map.Entry<String, Terms> termEntry,
             }
         }
 
+        //Set all existing notifications
+        for (RecommendationNotification recommendationNotification : notifications) {
+            recommendationModel.addNotification(recommendationNotification);
+        }
+
```
dinogun, line 1235: Is this the right place to add these notifications? Can you please paste an example JSON object that has this notification

msvinaykumar: The reason notifications aren't being set is because the function checks if the flag isRecommendedCPURequestAvailable is set to false. In such cases, these notifications are lost, despite the existing code intending to set a notification here.


https://privatebin.corp.redhat.com/?0252306990b9c2fc#5KA1424pqZhQbbogX7uD8ho36VJk5jngts69uU9F4jYV

msvinaykumar:
> @msvinaykumar - Tests for these notifications aren't present in the testsuite, new tests are required. You can check the issue above to see for which notification codes tests were added earlier.

done

chandrams (approved):
LGTM

dinogun (approved):
LGTM

PR #1170, opened 2024-04-20
Fix datasource for kruize_local experiments

dinogun:


src/main/java/com/autotune/analyzer/serviceObjects/Converters.java
```diff
@@ -73,6 +73,7 @@ public static KruizeObject convertCreateExperimentAPIObjToKruizeObject(CreateExp
                 kruizeObject.setClusterName(createExperimentAPIObject.getClusterName());
                 kruizeObject.setMode(createExperimentAPIObject.getMode());
                 kruizeObject.setPerformanceProfile(createExperimentAPIObject.getPerformanceProfile());
+                kruizeObject.setDataSource(createExperimentAPIObject.getDatasource());
```
khansaad, line 76: The datasource column that we have in the `kruize_experiments` table is of type `jsonb`. So, when this kruizeObject is saved in the DB, we either need to make changes in the table to make the column as Varchar or we need to add converter methods to save it as JSON else it will store as null only in the  `kruize_experiments` table.

dinogun: In the kruize_local case we are modifying it to Varchar. See this [file](https://github.com/kruize/autotune/blob/mvp_demo/migrations/kruize_local_ddl.sql)

khansaad: Oh right!. I missed this part. 
Looks fine then since these two are independent of each other.

src/main/java/com/autotune/analyzer/recommendations/engine/RecommendationEngine.java
```diff
@@ -225,8 +225,10 @@ public KruizeObject prepareRecommendations(int calCount) {
             // set the performance profile
             setPerformanceProfile(kruizeObject.getPerformanceProfile());
             // get the datasource
-            // TODO: temporarily setting the only supported datasource, this needs to fetched from the KruizeObject
-            String dataSource = KruizeConstants.SupportedDatasources.PROMETHEUS;
+            // TODO: If no data source given use KruizeDeploymentInfo.monitoring_agent / default datasource
+            String dataSource = kruizeObject.getDataSource();
```
khansaad, line 229: So, we are expecting the `datasource` to be passed as part of create_experiment as of now ?

dinogun: For local monitoring it is mandatory. Added code to fail local monitoring create experiment API if datasource not provided now

khansaad (approved):
LGTM!

PR #1183, opened 2024-04-24
fix to listRecommendations API

msvinaykumar:


src/main/java/com/autotune/analyzer/recommendations/engine/RecommendationEngine.java
```diff
@@ -91,6 +91,22 @@ private static int getNumPods(Map<Timestamp, IntervalResults> filteredResultsMap
         return (int) Math.ceil(max_pods_cpu);
     }
 
+    /**
+     * Populates the given map with Prometheus Query Language (PromQL) queries for various metrics.
+     *
+     * @param promQls The map to be populated with PromQL queries.
+     */
+    private static void getPromQls(Map<AnalyzerConstants.MetricName, String> promQls) {
```
dinogun, line 99: Are there any changes in this file? It all looks like whitespace changes to me?

dinogun: nm I see the change now

PR #1209, opened 2024-06-18
Updated ubi-minimal version to 8.10

chandrams:
Updated ubi-minimal version to 8.10

### DIFF
--- a/Dockerfile.autotune
+++ b/Dockerfile.autotune
@@ -48,7 +48,7 @@ RUN jlink --strip-debug --compress 2 --no-header-files --no-man-pages --module-p
 #            Runtime Docker Image
 ##########################################################
 # Use ubi-minimal as the base image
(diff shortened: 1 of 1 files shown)


Dockerfile.autotune
```diff
@@ -16,7 +16,7 @@
 ##########################################################
 #            Build Docker Image
 ##########################################################
-FROM registry.access.redhat.com/ubi8/ubi-minimal:8.9 as mvnbuild-jdk17
+FROM registry.access.redhat.com/ubi8/ubi-minimal:8.10 as mvnbuild-jdk17
```
dinogun, line 19: Needs to updated in both places, BUILD and RUNTIME

chandrams: Updated, thank you

dinogun (approved):
LGTM

PR #1216, opened 2024-06-27
Fix PromQL queries for local monitoring

shreyabiradar07:
This PR has the following changes:

- Removes conflicting namespace label selectors to be able to generate recommendations for workloads with `openshift-` as prefix
- Verified the updated query output by running queries and `local_monitoring` demo script on on `Openshift`


dinogun (approved):
LGTM

PR #1324, opened 2024-10-04
Add adapter for DeviceDetails

### LINKED ISSUE #1312: GPU MIG Right sizing recommendations by kruize
### Describe the feature 

Kruize reads CPU & memory usage data from the provided data source and comes up with the CPU and Memory right sizing recommendation. In a similar way it would be good to have the GPU MIG partition sizing recommendation for container which utilise GPU's

### Examples or references

Most of the ML workloads need GPU power and advanced GPU's from NVIDIA support MIG (Multi instance GPU's) where a single Physical GPU can be partitioned into multi instances of virtual or logical GPU's which can be configured and shared across multiple containers. Ampere (from A30) and Hopper series GPU's provide this feature.

### Suggest a solution

- Record the GPU related metrics
- Process the metrics along with CPU and Memory metrics
- Provide MIG partition recommendation

### Additional Context

None

bharathappali:
## Description

This PR is eighth part of the set of PR's which will be raised for adding GPU support #1312

This PR adds adapter for DeviceDetails to serialize and deserialize the class

This PR is build on top of #1321, so it needs to be merged after merging #1321

Fixes #1312 

### Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Docs update
- [ ] Requires DB changes

## How has this been tested?

Tests will be added in the other PR

- [ ] New Test X
- [ ] Functional testsuite

**Test Configuration**

## Checklist :dart:

- [ ] Comments added
- [ ] Documentation updated
- [ ] Tests added or updated

## Additional information

None

### DIFF
--- a/src/main/java/com/autotune/analyzer/adapters/DeviceDetailsAdapter.java
+++ b/src/main/java/com/autotune/analyzer/adapters/DeviceDetailsAdapter.java
@@ -8,6 +8,11 @@ import com.google.gson.stream.JsonReader;
 import com.google.gson.stream.JsonWriter;
 import java.io.IOException;
 
+
+/**
+ * This adapter actually specifies the GSON to identify the type of implementation of DeviceDetails
+ * to serialize or deserialize
+ */
 public class DeviceDetailsAdapter extends TypeAdapter<DeviceDetails> {
 
     @Override
--- a/src/main/java/com/autotune/analyzer/adapters/DeviceDetailsAdapter.java
+++ b/src/main/java/com/autotune/analyzer/adapters/DeviceDetailsAdapter.java
@@ -1,0 +1,79 @@
+package com.autotune.analyzer.adapters;
+
+import com.autotune.analyzer.utils.AnalyzerConstants;
+import com.autotune.common.data.system.info.device.DeviceDetails;
+import com.autotune.common.data.system.info.device.accelerator.AcceleratorDeviceData;
+import com.google.gson.TypeAdapter;
+import com.google.gson.stream.JsonReader;
+import com.google.gson.stream.JsonWriter;
+import java.io.IOException;
+
+public class DeviceDetailsAdapter extends TypeAdapter<DeviceDetails> {
+
+    @Override
+    public void write(JsonWriter out, DeviceDetails value) throws IOException {
+        out.beginObject();
+        out.name("type").value(value.getType().name());
+
+        if (value instanceof AcceleratorDeviceData accelerator) {
+            out.name("manufacturer").value(accelerator.getManufacturer());
+            out.name("modelName").value(accelerator.getModelName());
+            out.name("hostName").value(accelerator.getHostName());
+            out.name("UUID").value(accelerator.getUUID());
+            out.name("deviceName").value(accelerator.getDeviceName());
+            out.name("isMIG").value(accelerator.isMIG());
+        }
+        // Add for other devices when added
+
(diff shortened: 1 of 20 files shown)


dinogun:
Needs to be rebased against mvp_demo

bharathappali:
Apologies for the overlook, I have rebased to `mvp_demo`. Thanks @dinogun

src/main/java/com/autotune/analyzer/adapters/DeviceDetailsAdapter.java
```diff
@@ -0,0 +1,79 @@
...
+
+import com.autotune.analyzer.utils.AnalyzerConstants;
+import com.autotune.common.data.system.info.device.DeviceDetails;
+import com.autotune.common.data.system.info.device.accelerator.AcceleratorDeviceData;
+import com.google.gson.TypeAdapter;
+import com.google.gson.stream.JsonReader;
+import com.google.gson.stream.JsonWriter;
+import java.io.IOException;
+
+public class DeviceDetailsAdapter extends TypeAdapter<DeviceDetails> {
```
dinogun, line 11: javadoc

src/main/java/com/autotune/analyzer/adapters/RecommendationItemAdapter.java
```diff
@@ -0,0 +1,26 @@
+
+package com.autotune.analyzer.adapters;
+
+import com.autotune.analyzer.utils.AnalyzerConstants;
+import com.google.gson.*;
+
+import java.lang.reflect.Type;
+
+public class RecommendationItemAdapter implements JsonSerializer<AnalyzerConstants.RecommendationItem>, JsonDeserializer<AnalyzerConstants.RecommendationItem> {
```
dinogun, line 9: Why is this needed?

bharathappali: Earlier the `RecommendationItem` enum has only two entries `cpu` and `memory`. At the time if serialization (store in DB or return as JSON via API) java has handled the `toString` conversion and have converted them to `"cpu"` and `"memory"` strings. They are also keys in the recommendation (requests & limits)

But in case of NVIDIA the resources have `/` and `.` in their string representation of the MIG name. So we cannot add them as enums as is, So we had to create an entry which accepts a string and then the toString returns the string value of it.

if this adapter is not in place then at the time of deserailisation the string entries are converted to enum entries and vice versa in serialisation. For example if the entry is `NVIDIA_GPU_PARTITION_1_CORE_5GB("nvidia.com/mig-1g.5gb")`  then tostring of it will be `nvidia.com/mig-1g.5gb` which will not match the enum entry `NVIDIA_GPU_PARTITION_1_CORE_5GB`

Also to maintain consistency we changed the `cpu` to `CPU` so without the adapter the JSON will be generated with `CPU` as the key.

dinogun: I meant that this needs to be added as a javadoc ... methods with no doc makes it really hard for anybody new to understand why something is necessary

src/main/java/com/autotune/analyzer/adapters/DeviceDetailsAdapter.java
bharathappali: Done.

dinogun (approved):
LGTM

PR #1340, opened 2024-10-10
changed Up metrics to look for just Success response

msvinaykumar:
## Description

Kruize checks if the datasource is reachable by querying the following metric:
http://127.0.0.1:9090/api/v1/query?query=up.

Prometheus/Thanos typically returns the current status of the targets (such as services or jobs) it monitors. However, in this case, Thanos was unable to provide those details, even though the response status was 200 OK.

Kruize only needs to confirm that the response is 200 OK to proceed, and currently, it's not concerned with the actual content of the response. Therefore, the fix will focus on simply checking for a 200 OK status and moving forward based on that

`{
  "status": "success",
  "data": {
    "resultType": "vector",
    "result": [
      {
        "metric": {
          "__name__": "up",
          "instance": "localhost:9090",
          "job": "prometheus"
        },
        "value": [1696841450.278, "1"]
      },
      {
        "metric": {
          "__name__": "up",
          "instance": "localhost:9100",
          "job": "node"
        },
        "value": [1696841450.278, "1"]
      }
    ]
  }
}
`

Fixes # (issue)

### Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Docs update
- [ ] Requires DB changes

## How has this been tested?

- [ ] New Test X
- [ ] Functional testsuite

**Test Configuration**

## Checklist :dart:

- [ ] Comments added
- [ ] Documentation updated
- [ ] Tests added or updated

## Additional information

### DIFF
--- a/src/main/java/com/autotune/common/datasource/prometheus/PrometheusDataOperatorImpl.java
+++ b/src/main/java/com/autotune/common/datasource/prometheus/PrometheusDataOperatorImpl.java
@@ -117,16 +117,11 @@ public class PrometheusDataOperatorImpl extends DataSourceOperatorImpl {
 
             if (null == jsonObject) {
                 return null;
+            } else {
+                return "1";   //if it returns HTTP STATUS_OK  200
             }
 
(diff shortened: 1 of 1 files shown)


dinogun:
Why is this change needed and why are we removing the old check ... can you please add some description to this PR and also to the code

msvinaykumar:
> Why is this change needed and why are we removing the old check ... can you please add some description to this PR and also to the code

Done

dinogun (approved):
LGTM

PR #1411, opened 2024-12-05
2. Update Fabric8 Version to 7.0.0 and Add VerticalPodScaler Client [KRUIZE-VPA Integration]

shekhar316:
## Description

This pull request includes the following updates and fixes:

- Upgraded the `Fabric8 client` library to latest version `7.0.0`.
- Added the `VerticalPodScaler` client to support operations related to VPA.
- Resolved issues introduced by the `Fabric8` library upgrade.

Fixes # (issue)

1. Updated the code to use `genericKubernetesResources()` instead of the deprecated `customResource()` method. Additionally, `getAdditionalProperties()` is used to access the custom resource data dynamically.
2. `WatcherException` is now the correct exception type for handling errors related to watchers, as the method signature has been updated. 
3. `kubernetesClient.events()` is changed to `kubernetesClient.v1().events()` as access to resources like events is now scoped under API versions.
4. The change from `Watcher<String>` to `Watcher<GenericKubernetesResource>` is because now it is the preferred way in new version. 

### Type of change

- [x] Bug fix
- [x] New feature
- [ ] Docs update
- [ ] Requires DB changes

## How has this been tested?

Tested on both the ResourceHub cluster and the Kind cluster to verify any issues. 

**Test Configuration**
* Kubernetes clusters tested on: ResourceHub (OpenShift), Kind

## Checklist :dart:

- [x] Comments added
- [ ] Documentation updated
- [ ] Tests added or updated

## Additional information

NA

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -8,7 +8,7 @@
     <artifactId>autotune</artifactId>
     <version>0.2</version>
     <properties>
-        <fabric8-version>6.13.4</fabric8-version>
+        <fabric8-version>7.0.0</fabric8-version>
         <org-json-version>20240303</org-json-version>
         <jetty-version>12.0.12</jetty-version>
         <slf4j-version>2.17.1</slf4j-version>
--- a/pom.xml
+++ b/pom.xml
@@ -8,7 +8,7 @@
     <artifactId>autotune</artifactId>
     <version>0.2</version>
     <properties>
-        <fabric8-version>4.13.2</fabric8-version>
+        <fabric8-version>6.13.4</fabric8-version>
         <org-json-version>20240303</org-json-version>
         <jetty-version>12.0.12</jetty-version>
         <slf4j-version>2.17.1</slf4j-version>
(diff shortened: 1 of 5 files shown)


pom.xml
```diff
@@ -8,7 +8,7 @@
     <artifactId>autotune</artifactId>
     <version>0.2</version>
     <properties>
-        <fabric8-version>4.13.2</fabric8-version>
+        <fabric8-version>6.13.4</fabric8-version>
```
dinogun, line 11: Looks like `7.0.0` was released 3 days ago ... can we check if that works for us

shekhar316: Sure ...

shekhar316: @dinogun I have updated the version to `7.0.0` and tested the PoC code and the current code. In version 7.0.0, the structure has changed, and the `VerticalPodAutoscaler` has been moved to the `autoscaling` API group. I will keep this change in mind and update the VPA code accordingly in upcoming PRs. It should not cause any issues.

PR #1459, opened 2025-01-16
KAMS - 2 : Add instaslice helper and settings updater

bharathappali:
## Description

KAMS - Kruize Accelerator MIG-tuning Service

This PR adds the changes to add a helper util functionality which uses updated fabric8 package to interact with instaslice objects. Also adds the settings updater functionality to check if the GPU recommendations can be applied via Instaslice

Fixes # (issue)

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Docs update
- [ ] Requires DB changes

## How has this been tested?

Tests are yet to be done. Currently parked as the resources are not available.

- [ ] New Test X
- [ ] Functional testsuite

**Test Configuration**

## Checklist :dart:

- [x] Comments added
- [ ] Documentation updated
- [ ] Tests added or updated


src/main/java/com/autotune/Autotune.java
```diff
@@ -173,6 +174,9 @@ public static void main(String[] args) {
             startAutotuneNormalMode(context);
         }
 
+        // Check the settings initially while starting
+        SettingsUpdater.initialiseCentralSettings();
```
dinogun, line 178: Please give it a better name, CentralSettings doesn't elaborate what these settings are about

dinogun:
@bharathappali Can you please address the review comments

src/main/java/com/autotune/Autotune.java
bharathappali: How about `MainSettings`?

bharathappali:
> @bharathappali As we have discussed before, autotune/analyser/utils and autotune/utils provides a very broad scope and unless a method needs such a broad scope, we need to narrow the scope to only those packages that actually use them. Also do we really need a new settings folder?

@dinogun I'm more inclined to have the settings folder at autotune src level as these settings have broader outlook and might be utilized in various places. Things like checking if dependent programs are running (Like instaslice) might have higher scope and not just related to one particular module like recommendation. In future we will be having a separate module to handle role back where we might need this, also in case of local monitoring (if we have access to read objects but not auto apply them) we can have these utils to check the dependencies like instaslice to extract info about existing partitions which are currently not available in datasource.

src/main/java/com/autotune/analyzer/recommendations/autoscaler/instaslice/InstasliceHelper.java
```diff
@@ -0,0 +1,133 @@
...
+        AtomicReference<String> uuid = new AtomicReference<>(null);
+
+        List<GenericKubernetesResource> instasliceObjects = getInstasliceObjects();
+        ObjectMapper mapper = new ObjectMapper();
+
+        if (null != instasliceObjects && !instasliceObjects.isEmpty()) {
+            try {
+                instasliceObjects.forEach(item -> {
+                    String name = item.getMetadata().getName();
+                    System.out.println("Found Instaslice: " + name);
```
dinogun, line 76: Please use LOGGER.info where absolutely essential and LOGGER.debug in the rest of the places

bharathappali: Updated print statements to Logger.

src/main/java/com/autotune/analyzer/autoscaler/instaslice/InstasliceHelper.java
```diff
@@ -0,0 +1,137 @@
...
+                    LOGGER.debug(AnalyzerConstants.AutoscalerConstants.InfoMsgs.FOUND_INSTASLICE, name);
+
+                    Map<String, Object> allocations = (Map<String, Object>) ((Map<String, Object>) item.getAdditionalProperties().get("spec")).get("allocations");
+                    for (Map.Entry<String, Object> entry : allocations.entrySet()) {
+                        String key = entry.getKey();
+                        Map<String, Object> allocationMap = (Map<String, Object>) entry.getValue();
+                        if (allocationMap.get("namespace").toString().equalsIgnoreCase(workloadNamespace)
+                                && allocationMap.get("podName").toString().startsWith(workloadName)) {
+                            profile.set(allocationMap.get("profile").toString());
+                            System.out.println(allocationMap.get("profile"));
```
dinogun, line 124: Is this needed?

bharathappali: Thanks for pointing out, it's not needed. Have removed it.

dinogun (approved):
LGTM

PR #1512, opened 2025-02-20
Bump mvp_demo to Kruize v0.4

dinogun:
Also better resource defaults for Kruize on Openshift to help with the bulk_demo


chandrams (approved):
LGTM

PR #1529, opened 2025-03-10
[Kruize VPA Integration] Adding Support for Dry Run Check For Auto Scaler 

shekhar316:
## Description

This PR adds support for dry run API that returns an error when attempting to exceed the namespace quota or limit ranges. This PR will also add recommendation notifications with error messages to alert users about the issue.

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Docs update
- [ ] Requires DB changes

## How has this been tested?

On Resource Hub cluster. 

**Test Configuration**
* Kubernetes clusters tested on: Openshift

## Checklist :dart:

- [x] Comments added
- [ ] Documentation updated
- [ ] Tests added or updated


shekhar316:
Kruize Image with the changes - `quay.io/rh-ee-shesaxen/autotune:dryrun`

src/main/java/com/autotune/analyzer/autoscaler/validator/ResourceValidator.java
```diff
@@ -0,0 +1,113 @@
...
+     * This method performs a dry run of a pod with the given resource values to determine if they are available.
+     * @param containerName String the name of the container for which resource verification is performed
+     * @param image String the name of the image used for the container
+     * @param namespace String the namespace where workload is deployed
+     * @param cpuValue Quantity the CPU resource request to be verified
+     * @param memoryValue Quantity the memory resource request to be verified
+     */
+
+    public static void dryRunPod(String containerName, String image, String namespace,
+                                    Quantity cpuValue, Quantity memoryValue, ValidationOutputData validationOutputData) {
```
bharathappali, line 70: Thanks @shekhar316 for adding Dry Run support.

I feel the dry run feature will be used more extensively in future and there might be number of other params that will be added. So I feel instead of passing `cpuValue` and `memoryValue` seperately, Can we create a map with name `resourceMap` and pass it to the method and place the required `null` checks before accessing the quantities?

I accept that there will be a slight increase in the memory consumption due to this step but as it's a short lived object I feel we can keep the flow clear by having a map for resources rather than passing values individually. This also helps in minimizing the method signature changes if we add any other resources like accelerators in future.

Please let me know your thoughts on this?

src/main/java/com/autotune/analyzer/autoscaler/validator/ResourceValidator.java
```diff
@@ -0,0 +1,113 @@
...
+        try {
+            KubernetesClient client = new DefaultKubernetesClient();
```
bharathappali, line 72: The scope of `client` is in try block, it would be a good practice to `close` the client session once it's usage is completed.

Something like:
```
client.close() // Add this at the end of try block
```

Or if the client is `auto closable` (I feel the kubernetes client is autoclosable) you can create a `try-with-resources` block instead of a simple `try` block

You code can be refactored something like this:

```
try (KubernetesClient client = new DefaultKubernetesClient()) {
... // Use the client here
} // Client will be auto closed after the block
```

bharathappali (changes requested):

src/main/java/com/autotune/analyzer/autoscaler/validator/ResourceValidator.java
shekhar316: Thanks @bharathappali for the suggestion! 
I've updated the method to use a `Map<String, Quantity> resourceMap` instead of passing CPU and memory values individually. This should make the method more flexible for future enhancements, like supporting additional resource types. Also added the null checks before accessing the values.

src/main/java/com/autotune/analyzer/autoscaler/validator/ResourceValidator.java
shekhar316: Sure, since the `KubernetesClient` is autocloseable, I've updated the code to use a try-with-resources block.

bharathappali (approved):
LGTM

dinogun (approved):
LGTM

PR #1584, opened 2025-05-31
[DailyByte]: Avoid creating un-referenced objects in loop

### LINKED ISSUE #1583: Avoid redundant creation of IntervalResults objects in result conversion logic
In `convertUpdateResultsAPIObjToExperimentResultData()`, a new `IntervalResults` object is being created inside the loop for each metric. However, only the last created object is retained in the `resultsMap`, making all previously created IntervalResults objects redundant and subject to garbage collection.

This leads to unnecessary object creation and temporary memory pressure.

[Link for code](https://github.com/kruize/autotune/blob/mvp_demo/src/main/java/com/autotune/analyzer/serviceObjects/Converters.java#L285)

**Proposed Fix:**

Refactor the logic to:
    - Create the `IntervalResults` object once per container (per interval).
    - Populate it with the `metricResultsMap` after collecting all metrics.
    - Then store it in the `resultsMap`.

This reduces GC overhead and improves performance.

**Memory breakdown of `IntervalResults` object:**

Field | Type | Size (Compressed) | Size (Uncompressed) | Notes
-- | -- | -- | -- | --
Object Header | JVM internal | 12 bytes | 16 bytes | Includes mark word and class pointer
metricResultsMap | HashMap<MetricName, MetricResults> | 4 bytes | 8 bytes | Reference only
acceleratorMetricResultHashMap | HashMap<MetricName, AcceleratorResult> | 4 bytes | 8 bytes | Reference only
intervalStartTime | Timestamp | 4 bytes | 8 bytes | Reference only
intervalEndTime | Timestamp | 4 bytes | 8 bytes | Reference only
durationInMinutes | Double | 8 bytes | 8 bytes | Primitive (boxed if used externally)
duration_in_seconds | Double | 8 bytes | 8 bytes | Primitive
Total (without padding/alignment) | — | 44 bytes | 64 bytes | Minimum memory before alignment/padding

Actual heap allocation will be rounded up to 8 or 16 byte alignment, so practical sizes may be 48 bytes (compressed) and 64 or 72 bytes (uncompressed).

So the total temporary memory saving would be 48 bytes (As compressed refs are turned on by default in JVM) per request / per timestamp / per kubernetes object / per container / per metric

bharathappali:
## Description

Interval Results object is getting created inside the loop which set's the metric result hashmap
For each metric we are not only updating the map but also the reference in interval results,
So it's like the same map is being referenced in different interval result objects and the
object created in the final iteration will be stored in resultsMap, all the objects created earlier
will be dereferenced internally in JVM which will be removed in the next GC cycle, Object creation is
costly operation and maintaining it till GC to happen will impact the temporary memory consumption.
So this needs to be optimised in a way that per timestamp, we create interval results only once and
populate it with all the metrics which are added in the map and then we attach the interval results to resultsMap

Fixes #1583 

### Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Docs update
- [ ] Requires DB changes

## How has this been tested?

- [ ] New Test X
- [ ] Functional testsuite

**Test Configuration**

## Checklist :dart:

- [ ] Comments added
- [ ] Documentation updated
- [ ] Tests added or updated

## Additional information

### DIFF
--- a/src/main/java/com/autotune/analyzer/serviceObjects/Converters.java
+++ b/src/main/java/com/autotune/analyzer/serviceObjects/Converters.java
@@ -304,12 +304,12 @@ public class Converters {
                         metricsMap.put(AnalyzerConstants.MetricName.valueOf(metric.getName()), metric);
                         MetricResults metricResults = metric.getMetricResult();
                         metricResults.setName(metric.getName());
-                        IntervalResults intervalResults = new IntervalResults(updateResultsAPIObject.startTimestamp,
-                                updateResultsAPIObject.endTimestamp);
                         metricResultsHashMap.put(AnalyzerConstants.MetricName.valueOf(metric.getName()), metricResults);
-                        intervalResults.setMetricResultsMap(metricResultsHashMap);
-                        resultsMap.put(updateResultsAPIObject.getEndTimestamp(), intervalResults);
                     }
+                    IntervalResults intervalResults = new IntervalResults(updateResultsAPIObject.startTimestamp,
+                            updateResultsAPIObject.endTimestamp);
+                    intervalResults.setMetricResultsMap(metricResultsHashMap);
+                    resultsMap.put(updateResultsAPIObject.getEndTimestamp(), intervalResults);
                     containerData.setResults(resultsMap);
                     containerDataHashMap.put(containerData.getContainer_name(), containerData);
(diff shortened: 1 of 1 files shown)


khansaad (approved):
LGTM

PR #1589, opened 2025-06-04
[DailyByte]: Improve code readability and consistancy in accessing constants

### LINKED ISSUE #1588: Improve code readability via importing constants in a right way
In `Autotune.java` we are accessing the constants in `ServerContext.java` in two different ways.

1. We are accessing it via classname ([ServerContext.ROOT_CONTEXT](https://github.com/kruize/autotune/blob/master/src/main/java/com/autotune/Autotune.java#L103))
2. We are directing importing the value and referencing it ([KRUIZE_HTTP_THREAD_POOL_COUNT](https://github.com/kruize/autotune/blob/master/src/main/java/com/autotune/Autotune.java#L94))

For any dev who is new to this codebase might feel some inconsistency in the way we access the constants, the two usage patterns mentioned above have same performance as it's a compile time effort to link the constants and in runtime we don't see any difference. But when it comes to coding experience, I feel referring with the class name makes more sense as the dev will be aware of the origin of the constant.

So changing all direct imports to classname based imports will help in dev experience and code readability.

bharathappali:
## Description

Some constants are referred with static access via ServerContext class name
Some of them are accessed by impoting all contents of ServerContext
To improve code readability and give the developer a clear idea on where the constants are
created, we will be accessing the constants via classname rather than importing all the constants

Fixes #1588 

### Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Docs update
- [ ] Requires DB changes

## How has this been tested?

- [ ] New Test X
- [ ] Functional testsuite

**Test Configuration**

## Checklist :dart:

- [x] Comments added
- [ ] Documentation updated
- [ ] Tests added or updated


dinogun (approved):
LGTM

PR #1597, opened 2025-06-09
Add metric metadata structures

bharathappali:
## Description

This PR is built on top of #1596  it needs to be merged before this.

This PR adds the Metric Metadata structure to add the info sent via update results to kruize object

Fixes # (issue)

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Docs update
- [ ] Breaking change (What changes might users need to make in their application due to this PR?)
- [ ] Requires DB changes

## How has this been tested?

Please describe the tests that were run to verify your changes and steps to reproduce. Please specify any test configuration required. 

- [ ] New Test X
- [ ] Functional testsuite

**Test Configuration**
* Kubernetes clusters tested on: 

## Checklist :dart:

- [ ] Followed coding guidelines
- [ ] Comments added
- [ ] Dependent changes merged
- [ ] Documentation updated
- [ ] Tests added or updated

## Additional information

Include any additional information such as links, test results, screenshots here

PR #1598, opened 2025-06-09
Add metric metadata adapters

bharathappali:
## Description

This PR is built on top of #1597 

This PR add the Metric Metadata adapters to convert the JSON based results to java objects and to map them to appropriate implementation

Fixes # (issue)

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Docs update
- [ ] Requires DB changes

## How has this been tested?

- [ ] New Test X
- [ ] Functional testsuite

**Test Configuration**

## Checklist :dart:

- [ ] Comments added
- [ ] Documentation updated
- [ ] Tests added or updated

## Additional information


src/main/java/com/autotune/analyzer/adapters/MetricMetadataAdapter.java
```diff
@@ -0,0 +1,77 @@
...
+                    break;
+                default:
+                    in.skipValue();
+            }
+        }
+        in.endObject();
+
+        // For now, we assume that no type field present it's accelerator metadata
+        // In future if other metrics send different metadata then type cannot be null and type check should be mandatory
+        if (type == null || type.equalsIgnoreCase(AnalyzerConstants.DeviceType.ACCELERATOR.toString())) {
```
khansaad, line 118: This `type` here needs better handling, I feel. Even though we're assuming it to be either accelrator or null, there should be an proper error thrown for other cases.

src/main/java/com/autotune/analyzer/adapters/MetricMetadataAdapter.java
```diff
@@ -0,0 +1,77 @@
...
+            return new AcceleratorMetricMetadata(modelName, profileName, node);
+        }
+        return null;
+    }
+}
```
khansaad: EOF missing

src/main/java/com/autotune/analyzer/adapters/MetricMetadataAdapter.java
```diff
@@ -0,0 +1,77 @@
+package com.autotune.analyzer.adapters;
```
khansaad, line 16: Add copyright details

src/main/java/com/autotune/analyzer/adapters/MetricMetadataAdapter.java
```diff
@@ -0,0 +1,77 @@
...
+    public void write(JsonWriter out, MetricMetadata value) throws IOException {
+        if (value == null) {
+            out.nullValue();
+            return;
+        }
+        out.beginObject();
+        if (null != value.getType())
+            out.name(KruizeConstants.JSONKeys.TYPE).value(value.getType());
+
+        if (value instanceof AcceleratorMetricMetadata accel) {
```
khansaad, line 76: If I understand it correctly, for every new type, we would require modifications in this everytime ?

src/main/java/com/autotune/analyzer/adapters/MetricMetadataAdapter.java
```diff
@@ -0,0 +1,77 @@
...
+import com.google.gson.TypeAdapter;
+import com.google.gson.stream.JsonReader;
+import com.google.gson.stream.JsonWriter;
+import com.autotune.utils.KruizeConstants;
+
+
+import java.io.IOException;
+
+/**
+ * This adapter actually specifies the GSON to identify the type of implementation of MetricMetadata
```
khansaad: I think this javadoc needs to be more descriptive about current implementation and future prospects

khansaad (changes requested):

src/main/java/com/autotune/analyzer/adapters/MetricMetadataAdapter.java
bharathappali: Yes, for every implementation of `MetricMetadata` we need to add a modification. We should tell the adapter which fields to look for and populate in the implemented class object.

src/main/java/com/autotune/analyzer/adapters/MetricMetadataAdapter.java
bharathappali: Thanks for the review @khansaad. The reason for considering `null` as `Accelerator` is because right now we don't have any other metadata being passed via update results API. and sending `type` for every update result will increase the storage on db as it will be repetitive in all the records of GPU workloads.

And the reason for not throwing error is because there are null objects passed in case of CPU and Memory metrics where `MetricMetadata` is `null`  so in such cases we will mark the adpter conversion as null and skip the value

Even if we would like to avoid unsupported `type` we should have a guard placed at the update results api validation. We cannot raise exceptions in adapters as it will stop the conversion mid way and fails the whole conversion process. So it would be better to raise  exception before going to the GSON step.

Please let me know if this justifies the approach we took. Thanks in advance!

dinogun (approved):
LGTM

PR #1615, opened 2025-07-22
Update Maven distribution URL in the Dockerfile

khansaad:
## Description

This PR updates the maven distribution URL in the docker file to use the archive directory instead with the version `3.9.11`

Fixes # (issue)

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Docs update
- [ ] Requires DB changes

## How has this been tested?

Built the image from the scratch to see the correct maven version is being pulled and the build completes successfully 

- [ ] New Test X
- [ ] Functional testsuite

**Test Configuration**
* Kubernetes clusters tested on: Openshift

## Checklist :dart:

- [ ] Comments added
- [ ] Documentation updated
- [ ] Tests added or updated

## Additional information

Logs showing the change is working as expected

image (1)

### DIFF
--- a/Dockerfile.autotune
+++ b/Dockerfile.autotune
@@ -25,7 +25,7 @@ RUN microdnf --setopt=install_weak_deps=0 --setopt=tsflags=nodocs install -y jav
      && microdnf clean all 
 
 RUN mkdir -p /usr/share/maven /usr/share/maven/ref \
-  && curl -fsSL -o /tmp/apache-maven.tar.gz https://dlcdn.apache.org/maven/maven-3/3.9.11/binaries/apache-maven-3.9.11-bin.tar.gz \
+  && curl -fsSL -o /tmp/apache-maven.tar.gz https://archive.apache.org/dist/maven/maven-3/3.9.8/binaries/apache-maven-3.9.8-bin.tar.gz \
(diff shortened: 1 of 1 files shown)


Dockerfile.autotune
```diff
@@ -25,7 +25,7 @@ RUN microdnf --setopt=install_weak_deps=0 --setopt=tsflags=nodocs install -y jav
      && microdnf clean all 
 
 RUN mkdir -p /usr/share/maven /usr/share/maven/ref \
-  && curl -fsSL -o /tmp/apache-maven.tar.gz https://dlcdn.apache.org/maven/maven-3/3.9.11/binaries/apache-maven-3.9.11-bin.tar.gz \
+  && curl -fsSL -o /tmp/apache-maven.tar.gz https://archive.apache.org/dist/maven/maven-3/3.9.8/binaries/apache-maven-3.9.8-bin.tar.gz \
```
chandrams: I see 3.9.11 too, shouldn't we use that?

dinogun: @khansaad I suggested using archive so that we dont have links broken every few weeks, however we should use the latest one available now. So lets use 3.9.11 as suggested by Chandrakala

khansaad: Apologies, missed the updated version. 
Have added it now.

PR #205, opened 2021-03-11
Included a cron job for sanity tests

chandrams:
Signed-off-by: Chandrakala Subramanyam <csubrama@redhat.com>

PR #1654, opened 2025-09-29
Add Delete-Perf-Profile API

khansaad:
## Description

This PR enables the Delete Performance Profile API

Fixes # (issue)

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Docs update
- [ ] Requires DB changes

## How has this been tested?

- [ ] New Test X
- [ ] Functional testsuite

**Test Configuration**

## Checklist :dart:

- [ ] Comments added
- [ ] Documentation updated
- [ ] Tests added or updated

## Additional information

### DIFF
--- a/src/main/java/com/autotune/analyzer/services/PerformanceProfileService.java
+++ b/src/main/java/com/autotune/analyzer/services/PerformanceProfileService.java
@@ -234,10 +234,11 @@ public class PerformanceProfileService extends HttpServlet {
             }
             // Check if the profile is associated with any of the existing experiments
             // fetch experiments if any, associated with the mentioned profile name
-            List<String> experimentsList = new ExperimentDBService().loadExperimentNamesFromDBByProfileName(perfProfileName);
-            if (null != experimentsList && !experimentsList.isEmpty()) {
+            Long experimentsCount = new ExperimentDBService().getExperimentsCountFromDBByProfileName(perfProfileName);
+            if (experimentsCount != 0) {
                 sendErrorResponse(resp, null, HttpServletResponse.SC_BAD_REQUEST,
-                        String.format(AnalyzerErrorConstants.AutotuneObjectErrors.PERF_PROFILE_EXPERIMENTS_ERROR, experimentsList));
+                        String.format(AnalyzerErrorConstants.AutotuneObjectErrors.PERF_PROFILE_EXPERIMENTS_ERROR,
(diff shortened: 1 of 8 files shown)


khansaad:
> @khansaad Deleting a performance profile which still has entries in the `ExperimentTable` means that all of those experiments are going to throw an error. While deleting we need to check if it is still in use, then abort it. Only delete if it is already redundant

Handled it now. Thanks for catching this scenario.

src/main/java/com/autotune/analyzer/services/PerformanceProfileService.java
```diff
@@ -183,47 +187,83 @@ protected void doPut(HttpServletRequest req, HttpServletResponse resp) throws Se
...
+                sendErrorResponse(resp, null, HttpServletResponse.SC_BAD_REQUEST,
+                        AnalyzerErrorConstants.AutotuneObjectErrors.MISSING_PERF_PROFILE + perfProfileName);
+                return;
+            }
+            // Check if the profile is associated with any of the existing experiments
+            // fetch experiments if any, associated with the mentioned profile name
+            List<String> experimentsList = new ExperimentDBService().loadExperimentNamesFromDBByProfileName(perfProfileName);
+            if (null != experimentsList && !experimentsList.isEmpty()) {
+                sendErrorResponse(resp, null, HttpServletResponse.SC_BAD_REQUEST,
+                        String.format(AnalyzerErrorConstants.AutotuneObjectErrors.PERF_PROFILE_EXPERIMENTS_ERROR, experimentsList));
```
dinogun: This code looks dangerous! If we return all the experiments associated with a profile, as it stands in prod right now that will run into millions!! It will most definitely cause a OOM.

I'd suggest adding a new DB API that just returns the count of experiments using the Profile in Line #205 and print that number in the error message.

chandrams: @dinogun @khansaad Can we implement this later for the next release, as delete perf profile API is used only in the tests?

khansaad: @dinogun I have added the new API now. It returns the experiment count now:

```
{
    "message": "Performance Profile 'resource-optimization-openshift' cannot be deleted as it is currently associated with 1 experiment.",
    "httpcode": 400,
    "documentationLink": "",
    "status": "ERROR"
}
```

PR #1665, opened 2025-10-24
Simplified updatePerformanceProfile validation logic for improved compatibility

khansaad:
## Description

This PR eases the superset validation logic to allow modifications in the queries for the existing function variables.

**Reason**
We identified that the existing Performance Profile (v1) in the ROS repository failed the validations introduced in the updatePerformanceProfile API.

The issue occurred because the validation logic performed deep checks on the SLO data, including the query objects, and rejected the profile when query definitions differed.

During recent updates related to Namespace and GPU recommendations, some queries for existing function variables like `cpuRequest`, `cpuUsage`, etc were modified, which caused the validation to fail since it expected identical queries.

To address this, this PR relaxes the validation to only verify the presence and correctness of function variable names, without validating the full query structure.

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Docs update
- [ ] Requires DB changes

## How has this been tested?

- [ ] New Test X
- [ ] Functional testsuite

**Test Configuration**

## Checklist :dart:

- [ ] Comments added
- [ ] Documentation updated
- [ ] Tests added or updated

## Additional information

### DIFF
--- a/tests/scripts/remote_monitoring_tests/json_files/resource_optimization_openshift_v1.json
+++ b/tests/scripts/remote_monitoring_tests/json_files/resource_optimization_openshift_v1.json
@@ -17,11 +17,11 @@
         "aggregation_functions": [
           {
             "function": "avg",
-            "query": "avg by(container, namespace)(kube_pod_container_resource_requests{container!=\"\", container!=\"POD\", pod!=\"\", container=\"$CONTAINER_NAME$\", namespace=\"$NAMESPACE\", resource=\"cpu\", unit=\"core\"})"
+            "query": "avg(kube_pod_container_resource_requests{pod=~\"$DEPLOYMENT_NAME$-[^-]*-[^-]*$\", container=\"$CONTAINER_NAME$\", namespace=\"$NAMESPACE\", resource=\"cpu\", unit=\"core\"})"
           },
           {
             "function": "sum",
-            "query": "sum by(container, namespace)(kube_pod_container_resource_requests{container!=\"\", container!=\"POD\", pod!=\"\", container=\"$CONTAINER_NAME$\", namespace=\"$NAMESPACE\", resource=\"cpu\", unit=\"core\"})"
+            "query": "sum(kube_pod_container_resource_requests{pod=~\"$DEPLOYMENT_NAME$-[^-]*-[^-]*$\", container=\"$CONTAINER_NAME$\", namespace=\"$NAMESPACE\", resource=\"cpu\", unit=\"core\"})"
           }
         ]
       },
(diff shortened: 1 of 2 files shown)


bharathappali:
After looking at the changes in perf profile JSON, I can see that the major change is with `pod` attribute, earlier we are having it to match the regex having deployment name and it's changed to a simple query optimization of `pod!=""` which is also correct.

But as I can see we need to map the `v1` in production but not the one in the repo, we can revert the query changes for now [as we are not using the queries in remote monitoring usecase but maintaining them as reference for user while they collect the data and send it to kruize]

Also removing the field level validation will help here to avoid such issues, we can have them in place if needed in future. 

If the `pod` attribute optimisation is needed we can re-add them as part of `v2` which I feel will make it a bit clear. 

Thought? @khansaad 

cc: @dinogun

bharathappali:
After looking at the changes in perf profile JSON, I can see that the major change is with `pod` attribute, earlier we are having it to match the regex having deployment name and it's changed to a simple query optimization of `pod!=""`

But as I can see we need to map the `v1` in production but not the one in the repo, we can revert the query changes for now [as we are not using the queries in remote monitoring usecase but maintaining them as reference for user while they collect the data and send it to kruize]

Also removing the field level validation will help here to avoid such issues, we can have them in place if needed in future.

If the `pod` attribute optimization is needed we can re-add them as part of `v2` which I feel will make it a bit clear.

Thoughts? @khansaad

cc: @dinogun

bharathappali (approved):
To keep the test JSON in sync with the version that ROS has, These changes and validation logic looks good to me.

tests/scripts/remote_monitoring_tests/json_files/resource_optimization_openshift_v1.json
```diff
@@ -17,11 +17,11 @@
         "aggregation_functions": [
           {
             "function": "avg",
-            "query": "avg by(container, namespace)(kube_pod_container_resource_requests{container!=\"\", container!=\"POD\", pod!=\"\", container=\"$CONTAINER_NAME$\", namespace=\"$NAMESPACE\", resource=\"cpu\", unit=\"core\"})"
+            "query": "avg(kube_pod_container_resource_requests{pod=~\"$DEPLOYMENT_NAME$-[^-]*-[^-]*$\", container=\"$CONTAINER_NAME$\", namespace=\"$NAMESPACE\", resource=\"cpu\", unit=\"core\"})"
```
kusumachalasani, line 20: For `pod=~"$DEPLOYMENT_NAME$-[^-]*-[^-]*$"`
Can you clarify where the variable DEPLOYMENT_NAME is being used from? Is it referring to the workload name?

If yes, this regex expression doesn’t work for all workload types. It works correctly for workloads of type Deployment, but for StatefulSets, the naming pattern differs (for example, it follows $DEPLOYMENT_NAME$-[^-]*), so this expression may not match correctly in those cases.

chandrams:
@khansaad - I see the superset error in the perf profile migration test with this build - quay.io/dinogun210/autotune_operator:0.7.1-rc1, that uses the perf profile v1 from the testsuite initially, can you check this.

https://ci.app-svc-perf.corp.redhat.com/job/ExternalTeams/job/Autotune/job/kruize_perf_profile_migration_test/4/console

PR #1686, opened 2025-11-10
Container placeholder for layer detection

shreyabiradar07:
## Description

This PR supports `CONTAINER` placeholder in `hotspot-layer.json`, enabling dynamic layer detection for a given container experiment.

Fixes # (issue)

### Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Docs update
- [ ] Requires DB changes

## How has this been tested?

- [ ] New Test X
- [ ] Functional testsuite

**Test Configuration**

## Checklist :dart:

- [ ] Comments added
- [ ] Documentation updated
- [ ] Tests added or updated

## Additional information
docker image: `quay.io/shbirada/kruize_layer:layer_detection`

### DIFF
--- a/design/sample-jsons/layers/hotspot-layer.json
+++ b/design/sample-jsons/layers/hotspot-layer.json
@@ -11,7 +11,7 @@
       {
         "type": "query",
         "datasource": "prometheus-1",
-        "query": "jvm_memory_used_bytes{area=\"heap\",id=\"Eden Space\",container=\"$CONTAINER\"}",
+        "query": "jvm_memory_used_bytes{area=\"heap\",id=\"Eden Space\",container=\"$CONTAINER$\"}",
         "key": "pod",
         "non_null_is_present": true
       }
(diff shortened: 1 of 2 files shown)


design/sample-jsons/layers/hotspot-layer.json
```diff
@@ -11,7 +11,7 @@
       {
         "type": "query",
         "datasource": "prometheus-1",
-        "query": "jvm_memory_used_bytes{area=\"heap\",id=\"Eden Space\",container=\"tfb-server\"}",
+        "query": "jvm_memory_used_bytes{area=\"heap\",id=\"Eden Space\",container=\"$CONTAINER\"}",
```
bharathappali: Will it be a good idea to add the selector between two `$` where we can use regex in java code to find selectors which are in between two `$` rather than just starting with `$` as we cannot be sure about where the selector ends if there is no end-delimiter

bharathappali: so instead of `$CONTAINER` can we add `$CONTAINER$`

shreyabiradar07: done

bharathappali (approved):
LGTM

PR #1687, opened 2025-11-10
Update Runtimes Recommendation Logic

kusumachalasani:
## Description

Update Runtimes Recommendation Logic to include java version (Hard coded) details to come up with GC policy.


bharathappali (approved):
LGTM

PR #1720, opened 2025-12-08
Create Layer Configs for runtimes

kusumachalasani:
## Description

Create Layer Configs for runtimes. Includes the tunables gcpolicy, maxrampercentage for hotspot and openj9 ; threadpool.core-threads for quarkus.

### Type of change

- [ ] Bug fix
- [X] New feature
- [ ] Docs update
- [ ] Requires DB changes

### DIFF
--- a/manifests/autotune/layers/container-config.json
+++ b/manifests/autotune/layers/container-config.json
@@ -1,0 +1,30 @@
+{
+  "apiVersion": "recommender.com/v1",
+  "kind": "KruizeLayer",
+  "metadata": {
+    "name": "container"
+  },
+  "layer_name": "container",
+  "layer_level": 0,
+  "details": "generic container tunables",
+  "layer_presence": {
+    "presence": "always"
+  },
+  "tunables": [
+    {
+      "name": "memoryRequest",
+      "value_type": "double",
+      "lower_bound": "512 Mi",
+      "upper_bound": "2048 Mi",
+      "step": 1
+    },
+    {
+      "name": "cpuRequest",
+      "value_type": "double",
+      "lower_bound": "1.0",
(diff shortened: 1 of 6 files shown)


manifests/autotune/layers/container-config.json
```diff
@@ -0,0 +1,30 @@
+{
+  "apiVersion": "recommender.com/v1",
+  "kind": "KruizeLayer",
+  "metadata": {
+    "name": "container"
+  },
+  "layer_name": "container",
```
bharathappali, line 7: Thanks for the PR @kusumachalasani 

I see the `name` in `metadata` section, should we be maintaining again as `layer_name` or they are different entities and it's a coincidence for both to be same in this scenario?

manifests/autotune/layers/layer-config.yaml_template
```diff
@@ -30,18 +30,3 @@ tunables:
   upper_bound: '10'
```
chandrams, line 30: Is this template required? If this is required, we need some documentation briefing when this should be used.

manifests/autotune/layers/container-config.json
kusumachalasani: As the metadata name can be used for Kubernetes-related purposes, layer_name is meant for the application context. Even though we’re not using layer_name right now, it's helpful to retain it as it may be required in the future.

manifests/autotune/layers/layer-config.yaml_template
kusumachalasani: The template is for reference to create any new layer yaml. Will create some documentation explaining what the layer contains in a different PR as part of docs.

PR #1783, opened 2026-02-02
Updated the query to detect hotspot

chandrams:
## Description

Updated the query to detect hotspot layer

## Summary by Sourcery

Update hotspot layer sample JSON definition used for hotspot detection queries.


bharathappali (approved):
LGTM

PR #1790, opened 2026-02-10
Layer config updates

kusumachalasani:
## Description

Update the layer configs. This PR fixes
- memoryRequest tunable bounds format
- rename openj9 to semeru

Fixes # (issue)

### Type of change

- [X] Bug fix
- [ ] New feature
- [ ] Docs update
- [ ] Requires DB changes

## Summary by Sourcery

Update container layer configuration and add a new Semeru JVM layer definition.

New Features:
- Introduce a Semeru JVM layer configuration with heap presence detection queries and GC tuning options.

Bug Fixes:
- Correct memoryRequest tunable bounds format by removing unit suffixes from upper and lower bounds.

### DIFF
--- a/manifests/autotune/layers/hotspot-micrometer-config.json
+++ b/manifests/autotune/layers/hotspot-micrometer-config.json
@@ -60,7 +60,8 @@
       "value_type": "integer",
       "lower_bound": "25",
       "upper_bound": "90",
-      "step": 1
+      "step": 1,
+      "units": "%"
     }
   ]
 }
--- a/manifests/autotune/layers/hotspot-micrometer-config.yaml
+++ b/manifests/autotune/layers/hotspot-micrometer-config.yaml
@@ -43,4 +43,5 @@ tunables:
   lower_bound: '25'
   upper_bound: '90'
   step: 1
+  units: '%'
 
--- a/manifests/autotune/layers/semeru-actuator-config.json
+++ b/manifests/autotune/layers/semeru-actuator-config.json
@@ -55,7 +55,8 @@
       "value_type": "integer",
       "lower_bound": "25",
       "upper_bound": "90",
-      "step": 1
+      "step": 1,
+      "units": "%"
     }
   ]
 }
--- a/manifests/autotune/layers/semeru-actuator-config.yaml
+++ b/manifests/autotune/layers/semeru-actuator-config.yaml
@@ -40,4 +40,5 @@ tunables:
   lower_bound: '25'
   upper_bound: '90'
   step: 1
+  units: '%'
 


manifests/autotune/layers/container-config.json
```diff
@@ -14,8 +14,15 @@
     {
       "name": "memoryRequest",
       "value_type": "double",
-      "lower_bound": "512 Mi",
-      "upper_bound": "2048 Mi",
+      "lower_bound": "512",
+      "upper_bound": "2048",
```
dinogun, line 18: Why do we need this change? Is it not better to have the unit along with the number? So how is the unit going to be inferred if it is removed from here?

kusumachalasani: As we want to keep the tunables compatible with HPO, I removed the unit suffixes because HPO (Optuna) does not accept unit-based values when generating tunables from bound ranges.  If we have units, an extra unit-conversion layer in the tunable generation flow might be required.

Given this, I think it would be better to treat these values as bytes by default rather than Mi, since this also aligns with how memory metrics are retrieved from Prometheus. What do you think?

manifests/autotune/layers/container-config.yaml
```diff
@@ -9,14 +9,32 @@ layer_presence:
...
 tunables:
+# Memory request (MiB)
 - name: memoryRequest
   value_type: double
-  upper_bound: '2048 Mi'
-  lower_bound: '512 Mi'
+  upper_bound: '2048'
+  lower_bound: '512'
   step: 1
+  units: 'Mi'
```
bhanvimenghani, line 17: Should it be 'unit' singular or 'units' plural

manifests/autotune/layers/container-config.yaml
```diff
@@ -9,14 +9,32 @@ layer_presence:
...
+
+# Memory limit (MiB)
```
bhanvimenghani: maybe we can get rid of this comment as mib is mentioned in the units

kusumachalasani: Done.

manifests/autotune/layers/container-config.yaml
kusumachalasani: Industry standard in kubernetes is "units".

bhanvimenghani: Okay! Thanks

dinogun:
> > @kusumachalasani Will you be updating the .yaml files in a separate PR?
> 
> We introduced units as optional and have updated only for container layer files as that is the only file I felt more relevant to have units. Do you want me to introduce units for all tunables in all layers ?

Yes, I think it makes sense for all layers in general.

PR #1799, opened 2026-02-13
Layers - 11 Query Execution Validation at createlayer 

bhanvimenghani:
## Description
                                                                                                                                                                                                      
PromQL queries can fail in unpredictable ways depending on the Prometheus server configuration, available metrics, and query syntax. Since there is no standard PromQL query validation library, the most reliable approach is to execute queries against the actual datasources to ensure they work in the current deployment context.

This PR implements query execution validation during the createLayer API call. 

- It executes all queries against the first listed datasource, 
- validates result for the queries
- rejects layer creation if error response is received , 
- On success : logs validated queries

### Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Docs update
- [ ] Requires DB changes

## How has this been tested?

- [ ] New Test X
- [ ] Functional testsuite

**Test Configuration**

## Checklist :dart:

- [ ] Comments added
- [ ] Documentation updated
- [ ] Tests added or updated

## Additional information

## Summary by Sourcery

Validate PromQL-based layer queries during layer creation and simplify layer presence detection to be container-centric.

New Features:
- Execute and validate all query-based layer presence PromQL queries against all configured datasources during layer creation, rejecting layers when no datasource returns data.

Enhancements:
- Refine query-based presence detection to select datasources by provider type, log multi-datasource scenarios, and standardize logging for query validation and layer loading failures.
- Simplify the layer presence detection interface and utilities to operate on namespace and container only, deprecating workload-based detection and explicitly skipping unimplemented label-based presence detection.

### DIFF
--- a/src/main/java/com/autotune/analyzer/kruizeLayer/LayerValidation.java
+++ b/src/main/java/com/autotune/analyzer/kruizeLayer/LayerValidation.java
@@ -66,7 +66,7 @@ public class LayerValidation {
         }
 
         if (layer.getLayerPresence() == null) {
-            errors.add(AnalyzerErrorConstants.APIErrors.CreateLayerAPI.LAYER_PRESENCE_MISSING);
+            errors.add(AnalyzerErrorConstants.APIErrors.CreateLayerAPI.LAYER_PRESENCE_NULL);
         } else {
             // 2. Validate layer presence mutual exclusivity (only if not null)
             ValidationOutputData presenceValidation = validateLayerPresence(layer.getLayerPresence());


src/main/java/com/autotune/analyzer/kruizeLayer/utils/LayerUtils.java
```diff
@@ -66,7 +65,7 @@ public static Map<String, KruizeLayer> detectLayers(String containerName,
             experimentDBService.loadAllLayers(allLayersMap);
         } catch (Exception e) {
             LOGGER.error(LogMessages.FAILED_TO_LOAD_LAYERS, e);
-            throw new Exception("Failed to load layers from database: " + e.getMessage(), e);
+            throw new Exception(LogMessages.FAILED_TO_LOAD_LAYERS_EXCEPTION + ": " + e.getMessage(), e);
```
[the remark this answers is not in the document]
bhanvimenghani: done

src/main/java/com/autotune/analyzer/kruizeLayer/presence/QueryBasedPresence.java
```diff
@@ -185,8 +174,7 @@ private String appendFilter(String query, String labelName, String labelValue) {
         }
 
         // Escape the label value to prevent PromQL injection and syntax errors
-        String escapedValue = escapePromQLLabelValue(labelValue);
-        String filter = labelName + "=\"" + escapedValue + "\"";
+        String filter = labelName + "=\"" + labelValue.replace("\\", "\\\\").replace("\"","\\\"").replace("\n", "\\n").replace("\t", "\\t").replace("\r","\\r") + "\"";
```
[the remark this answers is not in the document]
bhanvimenghani: this is from the base pr 1794 i will address in that

src/main/java/com/autotune/analyzer/kruizeLayer/LayerValidation.java
```diff
@@ -178,4 +191,102 @@ private ValidationOutputData validateTunables(ArrayList<Tunable> tunables) {
...
+                    JsonArray resultArray = operator.getResultArrayForQuery(dataSourceInfo, baseQuery);
+
+                    // Check if this query returned data from this datasource
+                    if (resultArray != null && !resultArray.isEmpty()) {
+                        anyQueryReturnedData = true;
+                        datasourcesWithData.add(dataSourceInfo.getName());
+                        LOGGER.debug(LogMessages.QUERY_VALIDATION_SUCCESS_WITH_DATA,
+                                resultArray.size(), dataSourceInfo.getName());
+                    } else {
+                        LOGGER.debug(LogMessages.QUERY_VALIDATION_SUCCESS_NO_DATA,
```
bharathappali: @bhanvimenghani Thanks for raising the PR

I feel it would be good to know which particular query has returned empty / no data while debugging. Please kindly add the query name as well to the debug msg along with data source name.

bhanvimenghani: added query in the log msg

kusumachalasani (approved):
lgtm!

src/main/java/com/autotune/analyzer/kruizeLayer/LayerValidation.java
```diff
@@ -178,4 +191,102 @@ private ValidationOutputData validateTunables(ArrayList<Tunable> tunables) {
...
+            String baseQuery = query.getLayerPresenceQuery();
+            LOGGER.debug("Validating query: {}", baseQuery);
+
+            // Get all datasources
+            List<DataSourceInfo> allDatasources = new ArrayList<>(
+                    DataSourceCollection.getInstance().getDataSourcesCollection().values()
+            );
+
+            // Sort for deterministic order
+            allDatasources.sort(Comparator.comparing(DataSourceInfo::getName));
```
bharathappali: @bhanvimenghani I'm trying to understand the advantage we would be getting with sorting the list based on `name`. Please correct me if I'm wrong, I see that we are fetching the sources and iterating over them to find if query returns any data. So I feel we would be achieving this in any order, so I feel sorting is just adding up the time taken to execute this function, without adding any advantage.

bharathappali (changes requested):

src/main/java/com/autotune/analyzer/kruizeLayer/LayerValidation.java
bhanvimenghani: @bharathappali You're correct - the sorting was unnecessary overhead. We iterate through all datasources anyway and the order doesn't affect the validation outcome. Removed it. Thanks!

src/main/java/com/autotune/analyzer/kruizeLayer/LayerValidation.java
```diff
@@ -178,4 +191,97 @@ private ValidationOutputData validateTunables(ArrayList<Tunable> tunables) {
...
+
+            // Execute the base query without any filters
+            String baseQuery = query.getLayerPresenceQuery();
+            LOGGER.debug("Validating query: {}", baseQuery);
+
+            // Get all datasources and validate query against each
+            List<DataSourceInfo> allDatasources = new ArrayList<>(
+                    DataSourceCollection.getInstance().getDataSourcesCollection().values()
+            );
+            for (DataSourceInfo dataSourceInfo : allDatasources) {
```
dinogun: I'm thinking that this code is trying to do too much. The intent here is to catch any obvious user errors while adding the queries. So let us try to validate with the first valid datasource that matches the layer requirement.

Also dont need to calculate the datasources in the loop everytime. Move that out of the loop and ensure that the required datasource is used for each of the queries

bhanvimenghani: We added this logic to run queries against all the available datasources because in case of multiple datasources it can be possible that not all of them will have runtimes recommendation related information. (as in openshift usecase, we have 2 datasources and only one of them has runtime details)

PR #1807, opened 2026-02-17
Detect Quarkus layer through query

kusumachalasani:
## Description

Currently, Quarkus layer definition to detect the layer is through label. As label support is not implemented yet, converting the detection to query instead of label for now.

### DIFF
--- a/manifests/autotune/layers/quarkus-micrometer-config.json
+++ b/manifests/autotune/layers/quarkus-micrometer-config.json
@@ -10,7 +10,7 @@
     "queries": [
       {
         "datasource": "prometheus",
-        "query": "container_cpu_usage_seconds_total{} * on(namespace, pod) group_left(label_app_kubernetes_io_layer) kube_pod_labels{label_app_kubernetes_io_layer=\"quarkus\"}",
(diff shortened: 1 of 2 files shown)


manifests/autotune/layers/quarkus-micrometer-config.json
```diff
@@ -8,10 +8,11 @@
...
   "details": "quarkus tunables",
   "layer_presence": {
-    "label": [
+    "queries": [
       {
-        "name": "app.kubernetes.io/layer",
-        "value": "quarkus"
+        "datasource": "prometheus",
+        "query": "container_cpu_usage_seconds_total{} * on(namespace, pod) group_left(label_app_kubernetes_io_layer) kube_pod_labels{label_app_kubernetes_io_layer=\"quarkus\"}",
+        "key": "pod"
```
dinogun: Is this a standard label or something that we added? This is what is present in the [Quarkus](https://docs.redhat.com/en/documentation/red_hat_build_of_quarkus/1.11/html/release_notes_for_red_hat_build_of_quarkus_1.11/runtimes_metering_labels-quarkus-release-notes) documentation. I'd suggest that we use
```
com.redhat.component-name: "Quarkus"
```

kusumachalasani: Updated!

PR #1887, opened 2026-04-21
Adding changes to accomodate unified aggregrate query at metric level

afra-azf:
## Description

Please describe the issue or feature and the summary of changes made to fix this. 

Currently, multiple SQL queries are executed per metric attribute to compute aggregations like AVG, SUM, MIN, and MAX, leading to redundant scans.
This enhancement introduces a query attribute at the metric level to allow a single consolidated SQL query to return all required aggregation values, while still supporting the existing multi-query approach as a fallback when query is not provided.

Fixes # (issue)

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Docs update
- [ ] Requires DB changes

## How has this been tested?

**Tests Performed**
- Created a profile using unified aggregate query (single query returning avg/sum/min/max) with datasource db1.
- Used pre-existing data in ros_metrics_db (no new metric ingestion done).
- Identified a valid experiment from workloads table in ros_metrics_db.
- Created a Kruize experiment using the same workload details.
- Invoked:
     updateRecommendations
     listRecommendations

**Steps to Reproduce**
- Configure datasource db1 in dbconfigjson pointing to ros_metrics_db.
- Create a profile with unified aggregate SQL query.
- Pick an existing experiment from workloads table in ros_metrics_db.
- Create a Kruize experiment using same details.
- Fetch a valid interval_end_time from workload_metrics.
- Call:
    updateRecommendations
    listRecommendations

**Test Configuration**
- Existing dataset in ros_metrics_db:
     - workloads
     - workload_metrics
- No new data ingestion performed
- Kruize deployed on Kubernetes
- db1 configured as external datasource

**Validation **
- Unified aggregate query executed successfully against db1
- Data was fetched correctly from ros_metrics_db
- updateRecommendations generated expected outputs
- listRecommendations returned valid results
- No impact on existing flows or APIs

- [x] New Test : Unified aggregate query using external datasource (db1)
- [ ] Functional testsuite

**Test Configuration**
* Kubernetes clusters tested on: Kind
*  Kruize deployed in monitoring namespace
* External datasource: ros_metrics_db configured as db.
* Existing data used from:
    workloads
    workload_metrics

## Checklist :dart:

- [x] Comments added
- [ ] Documentation updated
- [ ] Tests added or updated

## Additional information

## Summary by Sourcery

Introduce support for unified aggregate SQL queries at the metric level while retaining the existing per-aggregation query behavior as a fallback.

New Features:
- Allow metrics to define a single unified SQL query that returns multiple aggregation values (e.g., avg, min, max, sum) per interval.
- Add support in the metrics DB connection manager to execute unified aggregate queries and map multiple result columns for a metric.

Enhancements:
- Extend recommendation engine metric fetching to use the unified aggregate query path when configured, falling back to individual aggregation queries for backward compatibility.
- Augment the Metric model with helpers to detect unified queries, derive expected result columns, and reuse query parameters from aggregation function definitions.

### DIFF
--- a/src/main/java/com/autotune/analyzer/recommendations/engine/RecommendationEngine.java
+++ b/src/main/java/com/autotune/analyzer/recommendations/engine/RecommendationEngine.java
@@ -1423,8 +1423,12 @@ public class RecommendationEngine implements RecommendationEngineService {
                                 queryParams.put("namespace", namespace);
                             else if ("experiment_name".equalsIgnoreCase(param))
                                 queryParams.put("experiment_name", experiment_name);
-                            else if ("start_time".equalsIgnoreCase(param))
-                                queryParams.put("start_time", experiment_name);
+                            else if ("start_time".equalsIgnoreCase(param)) {
+                            queryParams.put(
+                                "start_time",
+                                startTimeLocalDateTime.format(DateTimeFormatter.ISO_LOCAL_DATE_TIME)
+                            );
+                            }
                         }
                         
                         // Execute unified query
@@ -1460,8 +1464,8 @@ public class RecommendationEngine implements RecommendationEngineService {
                             for (String param : queryParamsFromProfile) {
                                 if ("container_name".equalsIgnoreCase(param))
                                     queryParams.put("container_name", container_name);
-                                else if ("namepsace".equalsIgnoreCase(param))
-                                    queryParams.put("namepsace", namespace);
+                                else if ("namespace".equalsIgnoreCase(param))
+                                    queryParams.put("namespace", namespace);
                                 else if ("experiment_name".equalsIgnoreCase(param))
                                     queryParams.put("experiment_name", experiment_name);
                                 else if ("start_time".equalsIgnoreCase(param))
@@ -1561,14 +1565,15 @@ public class RecommendationEngine implements RecommendationEngineService {
                         // Fallback to individual aggregation function queries (backward compatibility)
                         Set<String> aggrFunctions = metric.getAggregationFunctions();
                         for (String function : aggrFunctions) {
+                            queryParams.clear(); 
                             String query = metric.getQuery(function);
                             // Populate query params as defined in profile from standard list
                             List<String> queryParamsFromProfile = metric.getQueryParams(function);
                             for (String param : queryParamsFromProfile) {
                                 if ("container_name".equalsIgnoreCase(param))
                                     queryParams.put("container_name", container_name);
-                                else if ("namepsace".equalsIgnoreCase(param))
-                                    queryParams.put("namepsace", namespace);
+                                else if ("namespace".equalsIgnoreCase(param))
+                                    queryParams.put("namespace", namespace);
                                 else if ("experiment_name".equalsIgnoreCase(param))
                                     queryParams.put("experiment_name", experiment_name);
                                 else if ("start_time".equalsIgnoreCase(param))
@@ -1590,9 +1595,6 @@ public class RecommendationEngine implements RecommendationEngineService {
                                 intervalResults.setIntervalEndTime(interval_end_time);
                                 intervalResults.addMetricResult(metricName, function, value);
                             }
-                        }
-                    }
-                            //LOGGER.error("interval_start_time = {}, interval_end_time = {}, metricName = {}, function = {}, value = {}", interval_start_time, interval_end_time, metricName, function, value);
                         }
                     }
                 }
--- a/src/main/java/com/autotune/analyzer/recommendations/engine/RecommendationEngine.java
+++ b/src/main/java/com/autotune/analyzer/recommendations/engine/RecommendationEngine.java
@@ -1406,29 +1406,77 @@ public class RecommendationEngine implements RecommendationEngineService {
                     HashMap<Metric, MetricResults>  metricResults = new HashMap<>();
                     String datasourceName = metric.getDatasource();
                     MetricsDBConnectionManager metricsDBConnectionManager = MetricsDBConnectionManager.getInstance();
-                    Set<String> aggrFunctions = metric.getAggregationFunctions();
-                    for (String function : aggrFunctions) {
-                        String query = metric.getQuery(function).toLowerCase();
-                        List<String> queryParamsFromProfile = metric.getQueryParams(function);
+                    
+                    // Check if metric has unified query at metric level
+                    if (metric.hasUnifiedQuery()) {
+                        // Use unified query path - single query returns all aggregate values
+                        String unifiedQuery = metric.getQuery();
+                        List<String> queryParamsFromProfile = metric.getUnifiedQueryParams();
+                        List<String> expectedFunctions = metric.getExpectedResultColumns();
+                        
+                        // Populate query parameters
+                        queryParams.clear();
                         for (String param : queryParamsFromProfile) {
                             if ("container_name".equalsIgnoreCase(param))
                                 queryParams.put("container_name", container_name);
-                            else if ("namepsace".equalsIgnoreCase(param))
-                                queryParams.put("namepsace", namespace);
+                            else if ("namespace".equalsIgnoreCase(param))
+                                queryParams.put("namespace", namespace);
                             else if ("experiment_name".equalsIgnoreCase(param))
                                 queryParams.put("experiment_name", experiment_name);
                             else if ("start_time".equalsIgnoreCase(param))
                                 queryParams.put("start_time", experiment_name);
-
                         }
+                        
+                        // Execute unified query
+                        List<Object[]> unifiedResults = metricsDBConnectionManager.getUnifiedMetricsData(
+                            datasourceName, unifiedQuery, queryParams, expectedFunctions);
+                        
+                        if (unifiedResults != null && !unifiedResults.isEmpty()) {
+                            // Process unified results - each row contains all aggregate values
+                            for (Object[] row : unifiedResults) {
+                                Timestamp interval_start_time = (Timestamp) row[0];
+                                Timestamp interval_end_time = (Timestamp) row[1];
+                                
+                                IntervalResults intervalResults = results.computeIfAbsent(
+                                    interval_end_time, k -> new IntervalResults());
+                                
+                                // Process each aggregate function value
+                                for (int i = 0; i < expectedFunctions.size(); i++) {
+                                    String function = expectedFunctions.get(i);
+                                    String valueStr = (String) row[2 + i];
+                                    if (valueStr != null) {
+                                        Double value = Double.valueOf(valueStr);
+                                        intervalResults.addMetricResult(metricName, function, value);
+                                    }
+                                }
+                            }
+                        }
+                    } else {
+                        // Fallback to individual aggregation function queries (backward compatibility)
+                        Set<String> aggrFunctions = metric.getAggregationFunctions();
+                        for (String function : aggrFunctions) {
+                            String query = metric.getQuery(function).toLowerCase();
+                            List<String> queryParamsFromProfile = metric.getQueryParams(function);
+                            for (String param : queryParamsFromProfile) {
+                                if ("container_name".equalsIgnoreCase(param))
+                                    queryParams.put("container_name", container_name);
+                                else if ("namepsace".equalsIgnoreCase(param))
+                                    queryParams.put("namepsace", namespace);
+                                else if ("experiment_name".equalsIgnoreCase(param))
+                                    queryParams.put("experiment_name", experiment_name);
+                                else if ("start_time".equalsIgnoreCase(param))
+                                    queryParams.put("start_time", experiment_name);
 
-                        List<Object[]> dbResult = metricsDBConnectionManager.getMetricsData(datasourceName, query, queryParams);
-                        for (Object[] dbRow : dbResult) {
-                            Timestamp interval_start_time = (Timestamp) dbRow[0];
-                            Timestamp interval_end_time = (Timestamp) dbRow[1];
-                            Double value =  Double.valueOf((String) dbRow[2]);
-                            IntervalResults intervalResults = results.computeIfAbsent(interval_end_time, k -> new IntervalResults());
-                            intervalResults.addMetricResult(metricName, function, value);
+                            }
+
+                            List<Object[]> dbResult = metricsDBConnectionManager.getMetricsData(datasourceName, query, queryParams);
+                            for (Object[] dbRow : dbResult) {
+                                Timestamp interval_start_time = (Timestamp) dbRow[0];
+                                Timestamp interval_end_time = (Timestamp) dbRow[1];
+                                Double value =  Double.valueOf((String) dbRow[2]);
+                                IntervalResults intervalResults = results.computeIfAbsent(interval_end_time, k -> new IntervalResults());
+                                intervalResults.addMetricResult(metricName, function, value);
+                            }
                         }
                     }
                 }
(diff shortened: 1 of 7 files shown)


src/main/java/com/autotune/analyzer/recommendations/engine/RecommendationEngine.java
```diff
@@ -1458,37 +1512,91 @@ private void fetchContainerMetricsBasedOnProfile(KruizeObject kruizeObject, Time
...
                     String datasourceName = metric.getDatasource();
                     MetricsDBConnectionManager metricsDBConnectionManager = MetricsDBConnectionManager.getInstance();
-                    Set<String> aggrFunctions = metric.getAggregationFunctions();
-                    for (String function : aggrFunctions) {
-                        String query = metric.getQuery(function);
-                        // Populate query params as defined in profile from standard list
-                        List<String> queryParamsFromProfile = metric.getQueryParams(function);
+                    
+                    // Check if metric has unified query at metric level
+                    if (metric.hasUnifiedQuery()) {
```
mbvreddy, line 1517: We don't need another helper function here... 

```
String query = metric.getQuery();
if (query!= null) {
...
...
}
```

src/main/java/com/autotune/analyzer/recommendations/engine/RecommendationEngine.java
```diff
@@ -1458,37 +1512,91 @@ private void fetchContainerMetricsBasedOnProfile(KruizeObject kruizeObject, Time
...
+                        // Use unified query path - single query returns all aggregate values
+                        String unifiedQuery = metric.getQuery();
+                        List<String> queryParamsFromProfile = metric.getUnifiedQueryParams();
```
mbvreddy, line 1520: query_param attributes in in aggregation_functions[n] is for query defined for each function. And, query we define at metric level as part of this PR has different query params. So, you need to add `query_params` attribute for query at metric level.

src/main/java/com/autotune/analyzer/recommendations/engine/RecommendationEngine.java
```diff
@@ -1406,29 +1408,81 @@ private void fetchNamespaceMetricsBasedOnProfile(KruizeObject kruizeObject, Time
...
-                    for (String function : aggrFunctions) {
-                        String query = metric.getQuery(function).toLowerCase();
-                        List<String> queryParamsFromProfile = metric.getQueryParams(function);
+                    
+                    // Check if metric has unified query at metric level
+                    if (metric.hasUnifiedQuery()) {
+                        // Use unified query path - single query returns all aggregate values
+                        String unifiedQuery = metric.getQuery();
+                        List<String> queryParamsFromProfile = metric.getUnifiedQueryParams();
+                        List<String> expectedFunctions = metric.getExpectedResultColumns();
```
mbvreddy, line 1417: If query at metric level is declared, it is not mandatory to define `aggregation_functions`. We may not have these mentioned always. But, the thought of figuring out the columns is valid. So, like query_params, we may need to see if explicit declaration of result columns is required.

src/main/java/com/autotune/analyzer/recommendations/engine/RecommendationEngine.java
```diff
@@ -1458,37 +1512,91 @@ private void fetchContainerMetricsBasedOnProfile(KruizeObject kruizeObject, Time
...
-                        String query = metric.getQuery(function);
-                        // Populate query params as defined in profile from standard list
-                        List<String> queryParamsFromProfile = metric.getQueryParams(function);
+                    
+                    // Check if metric has unified query at metric level
+                    if (metric.hasUnifiedQuery()) {
+                        // Use unified query path - single query returns all aggregate values
+                        String unifiedQuery = metric.getQuery();
+                        List<String> queryParamsFromProfile = metric.getUnifiedQueryParams();
+                        List<String> expectedFunctions = metric.getExpectedResultColumns();
```
mbvreddy, line 1521: if query at metric level is declared, it is not mandatory to define aggregation_functions. We may not have these mentioned always. But, the thought of figuring out the columns is valid. So, like query_params, we may need to see if explicit declaration of result columns is required.

src/main/java/com/autotune/analyzer/recommendations/engine/RecommendationEngine.java
```diff
@@ -1458,37 +1512,91 @@ private void fetchContainerMetricsBasedOnProfile(KruizeObject kruizeObject, Time
...
                             else if ("end_time".equalsIgnoreCase(param))
                                 queryParams.put("end_time", endTimeLocalDateTime.format(DateTimeFormatter.ISO_LOCAL_DATE_TIME));
                         }
-                        //LOGGER.error("query = {}", query);
-                        List<Object[]> dbResult = metricsDBConnectionManager.getMetricsData(datasourceName, query, queryParams);
-                        if (dbResult == null) {
-                            return;
+                        
+                        // Execute unified query
+                        List<Object[]> unifiedResults = metricsDBConnectionManager.getUnifiedMetricsData(
```
mbvreddy, line 1539: we don't need another method. Use existing method and it is just that the query is diffrent.

src/main/java/com/autotune/analyzer/recommendations/engine/RecommendationEngine.java
```diff
@@ -1458,37 +1512,91 @@ private void fetchContainerMetricsBasedOnProfile(KruizeObject kruizeObject, Time
...
+                            
+                            IntervalResults intervalResults = results.computeIfAbsent(
+                                interval_end_time, k -> new IntervalResults(interval_start_time, interval_end_time));
                             intervalResults.setIntervalStartTime(interval_start_time);
                             intervalResults.setIntervalEndTime(interval_end_time);
-                            intervalResults.addMetricResult(metricName, function, value);
-                            //LOGGER.error("interval_start_time = {}, interval_end_time = {}, metricName = {}, function = {}, value = {}", interval_start_time, interval_end_time, metricName, function, value);
+                            
+                            // Process each aggregate function value
+                            for (int i = 0; i < expectedFunctions.size(); i++) {
```
mbvreddy, line 1557: Need to figure out better way here.

src/main/java/com/autotune/analyzer/recommendations/engine/RecommendationEngine.java
```diff
@@ -1458,37 +1512,91 @@ private void fetchContainerMetricsBasedOnProfile(KruizeObject kruizeObject, Time
...
+                                String function = expectedFunctions.get(i);
+                                String valueStr = (String) row[2 + i];
+                                if (valueStr != null) {
+                                    Double value = Double.valueOf(valueStr);
+                                    intervalResults.addMetricResult(metricName, function, value);
+                                }
+                            }
+                        }
+                    } else {
```
mbvreddy, line 1566: We might need to execute this logic even when query at metric level is defined and it doesn't return any data. So, `else` is not right fit.

src/main/java/com/autotune/analyzer/recommendations/engine/RecommendationEngine.java
```diff
@@ -1458,37 +1512,91 @@ private void fetchContainerMetricsBasedOnProfile(KruizeObject kruizeObject, Time
...
+                            List<String> queryParamsFromProfile = metric.getQueryParams(function);
+                            for (String param : queryParamsFromProfile) {
+                                if ("container_name".equalsIgnoreCase(param))
+                                    queryParams.put("container_name", container_name);
+                                else if ("namespace".equalsIgnoreCase(param))
+                                    queryParams.put("namespace", namespace);
+                                else if ("experiment_name".equalsIgnoreCase(param))
+                                    queryParams.put("experiment_name", experiment_name);
+                                else if ("start_time".equalsIgnoreCase(param))
+                                    queryParams.put("start_time", startTimeLocalDateTime.format(DateTimeFormatter.ISO_LOCAL_DATE_TIME));
```
mbvreddy, line 1582: This is repeated for query at metric level and query at function level. So, make it as a private helper method.

src/main/java/com/autotune/common/data/metrics/Metric.java
```diff
@@ -143,4 +143,31 @@ public Set<String> getAggregationFunctions() {
...
+
+    // Helper to get query_params from aggregation_functions 
+    public List<String> getUnifiedQueryParams() {
+        if (aggregationFunctionsMap == null || aggregationFunctionsMap.isEmpty()) {
+            return Collections.emptyList();
+        }
+        // Get query_params from first aggregation function
+        AggregationFunctions firstFunc = aggregationFunctionsMap.values().iterator().next();
+        return firstFunc.getQueryParams();
+    }
```
mbvreddy, line 172: As mentioned in previous comments, these helper methods are not required.

src/main/java/com/autotune/database/init/MetricsDBConnectionManager.java
```diff
@@ -249,6 +252,56 @@ public List<Object[]> getMetricsData(String metricsDbRef, String sql, Map<String
...
+    } finally {
+        if (session != null) {
+            try {
+                session.close();
+            } catch (Exception e) {
+                LOGGER.error("Error closing session", e);
+            }
+        }
+    }
+}
```
mbvreddy, line 304: Same. Reuse existing method. Only change is the query passed as argument.Other than that, there is no difference.

src/main/java/com/autotune/database/init/MetricsDBConnectionManager.java
```diff
@@ -249,6 +252,56 @@ public List<Object[]> getMetricsData(String metricsDbRef, String sql, Map<String
...
+        for (String function : expectedFunctions) {
+            query.addScalar(function + "_value", String.class);
+        }
+
+        if (params != null) {
+        for (String key : params.keySet()) {
+            if (!sql.contains(":" + key)) {
+                LOGGER.debug("Skipping unused param: {}", key);
+                continue;
+            }
```
mbvreddy, line 279: I don't think, we need this explicit validation here.

mbvreddy (changes requested):

src/main/java/com/autotune/analyzer/recommendations/engine/RecommendationEngine.java
afra-azf: We currently use positional mapping (row[2+i]) aligned with expected_result_columns, which is lightweight and backward compatible but depends on column order.
An alternative is alias-based mapping (using column names like AVG(cpu) AS avg and fetching via rowMap.get("avg")), which is more robust and readable but requires changes in query configs and DB layer. This may affect the backward compatibility. 
Please let me know which approach to go ahead with.

src/main/java/com/autotune/analyzer/recommendations/engine/RecommendationEngine.java
afra-azf: Removed the helper method and switched to directly checking metric.getQuery() != null to determine unified query execution, as suggested. This simplifies the flow and avoids unnecessary abstraction.

src/main/java/com/autotune/analyzer/recommendations/engine/RecommendationEngine.java
afra-azf: Added query_params at the metric (unified query) level as suggested. Unified queries now use metric.getQueryParams() instead of aggregation-level params, ensuring clear separation and correct parameter binding for different query types.

src/main/java/com/autotune/common/data/metrics/Metric.java
afra-azf: Introduced result_columns at metric level to explicitly define output columns for both unified queries and aggregate functions independently . This removes dependency on aggregation_functions and ensures unified queries can operate independently with clear, config-driven result mapping.

src/main/java/com/autotune/analyzer/recommendations/engine/RecommendationEngine.java
afra-azf: Introduced result_columns at metric level to explicitly define output columns for both unified queries and aggregate functions independently. This removes dependency on aggregation_functions and ensures unified queries can operate independently with clear, config-driven result mapping.

src/main/java/com/autotune/analyzer/recommendations/engine/RecommendationEngine.java
afra-azf: Introduced result_columns at metric level to explicitly define output columns for both unified queries and aggregate functions independently. This removes dependency on aggregation_functions and ensures unified queries can operate independently with clear, config-driven result mapping.

Although, as per existing functionality , aggregate functions are mandatory to be maintained during metric profile creation, in future we can remove the validation.

src/main/java/com/autotune/analyzer/recommendations/engine/RecommendationEngine.java
afra-azf: Removed the separate getUnifiedMetricsData method and reused the existing getMetricsData by passing the unified query and result_columns. This avoids duplication and keeps query execution logic consistent across unified and aggregation paths.

src/main/java/com/autotune/analyzer/recommendations/engine/RecommendationEngine.java
afra-azf: Flow chart of this approach:

https://ibm-my.sharepoint.com/:t:/p/afrafayaz/IQCw5mD8SclBSLq_hMluZ1PKAc25sMXvYtb3pygNBC9813s?e=aQHYrI

[](https://ibm-my.sharepoint.com/:t:/p/afrafayaz/IQCw5mD8SclBSLq_hMluZ1PKAc25sMXvYtb3pygNBC9813s?e=aQHYrI)

Replaced positional indexing (row[2 + i]) with a config-driven approach using result_columns and a columnIndexMap. While mapping is driven by column names, it still relies on maintaining consistent ordering between the SQL SELECT clause and result_columns. A fully name-based mapping approach (e.g., Tuple/Map) was considered but would require significant refactoring and introduce overhead, so this approach provides a balanced and performant solution.

**Why this approach was chosen:**

- Hibernate native queries return Object[] (positional results) by default
- Moving to true name-based mapping (e.g., Tuple / Map transformers) would:
- Require significant refactoring across DB layer and processing logic
- Introduce additional overhead and complexity
- Reduce performance compared to direct Object[] access

**Current approach provides a balanced solution:**

- Eliminates hardcoded indices 
- Keeps performance intact 
- Maintains flexibility via configuration

src/main/java/com/autotune/analyzer/recommendations/engine/RecommendationEngine.java
afra-azf: Removed the else condition and introduced a dataFound flag to control fallback execution. Aggregation logic is now triggered when unified query returns no data, ensuring correct behavior even when metric-level query is present but yields empty results.

src/main/java/com/autotune/analyzer/recommendations/engine/RecommendationEngine.java
afra-azf: Extracted query parameter population logic into a private helper method (populateQueryParams() )and reused it for both unified and aggregation queries. This removes duplication and ensures consistent parameter handling across execution paths.

src/main/java/com/autotune/database/init/MetricsDBConnectionManager.java
afra-azf: Updated approach :

List<Object[]> results = metricsDBConnectionManager.getMetricsData(
        datasource, query, queryParams, resultColumns);
        
  Removed the separate method and reused the existing getMetricsData by passing the query dynamically. Since only the query differs, this keeps the implementation simple and avoids duplication.

src/main/java/com/autotune/database/init/MetricsDBConnectionManager.java
afra-azf: Removed the explicit SQL parameter presence check and now directly set parameters on the query. This simplifies the implementation and aligns with existing query execution behavior.

mbvreddy (approved):

Issue #1977, opened 2026-06-23
Feature Request: Bulk Profile API for Automated Recommendation Generation


bharathappali:
### Description

Add a Bulk Profile API to enable reusable configuration profiles for automated, recurring bulk job execution. Users can define templates that automatically trigger bulk recommendation generation on a scheduled basis, eliminating manual job triggering.

##### Current (Manual): User must trigger bulk jobs daily

```bash
curl -X POST http://localhost:8080/bulk -d '{...}'  # Repeat daily
```

##### With Bulk Profile (Automated): Create once, runs automatically [via optimizer]

```bash
curl -X POST http://localhost:8080/bulkProfiles -d '{
  "profile_name": "prod-monitoring",
  "clusters": [{
    "cluster_name": "production",
    "datasources": ["prometheus-1"],
    "experiment_types": ["container"]
  }],
  "recommendation_settings": {
    "scheduling": "24h",
    "terms": ["short_term", "medium_term", "long_term"],
    "models": ["performance", "cost"]
  }
}'
```

### Task List

- [ ] [Create migration `v111__create_kruize_bulk_profile_table.sql` with table and indexes](https://github.com/kruize/autotune/pull/1978)
- [ ]  [Create JPA entity `KruizeBulkProfileEntry.java` with JSONB columns](https://github.com/kruize/autotune/pull/1979)
- [ ] [Add DAO methods in `ExperimentDAO` and `ExperimentDAOImpl` for CRUD operations](https://github.com/kruize/autotune/pull/1988)
- [ ] [Add HQL query constants to `DBConstants.java`](https://github.com/kruize/autotune/pull/1988)
- [ ] [Register entity in `KruizeHibernateUtil.buildSessionFactory()`](https://github.com/kruize/autotune/pull/1991)
- [ ] [Create `BulkProfile.java` with nested classes for cluster config and recommendation settings](https://github.com/kruize/autotune/pull/1980)
- [ ] [Create converters in `KruizeBulkProfileEntry.java` for to and fro conversion to `BulkProfile.java`](https://github.com/kruize/autotune/pull/1981)
- [ ] [Create `BulkProfileUpdateRequest.java` for partial updates](https://github.com/kruize/autotune/pull/1992)
- [ ] [Create `BulkProfileService.java` servlet with POST/GET/PUT/DELETE endpoints](https://github.com/kruize/autotune/pull/1994)
[...]

PR #1978, opened 2026-06-23
[BULK-PROFILE]-1: Add the bulk profile table

bharathappali:
## Description

Add the bulk profile table

Fixes #1977 

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Docs update
- [ ] Requires DB changes

## How has this been tested?

- [ ] New Test X
- [ ] Functional testsuite

**Test Configuration**

## Checklist :dart:

- [ ] Comments added
- [ ] Documentation updated
- [ ] Tests added or updated

## Additional information

## Summary by Sourcery

New Features:
- Add kruize_bulk_profile table to persist bulk profile metadata, associated clusters, and recommendation settings.


migrations/lm/v112__create_optimizer_bulk_config_table.sql
```diff
@@ -0,0 +1,9 @@
+CREATE TABLE IF NOT EXISTS kruize_bulk_profile (
+    profile_name VARCHAR(255) PRIMARY KEY,
+    clusters JSONB NOT NULL,
+    recommendation_settings JSONB NOT NULL,
+    webhook_url VARCHAR(500),
+    enabled BOOLEAN DEFAULT true,
+    created_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP,
+    updated_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP
```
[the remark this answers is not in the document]
bharathappali: we will ensure that `updated_at` field is updated on every update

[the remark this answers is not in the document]
bharathappali: updated the sql file

kusumachalasani (approved):
lgtm!

dinogun:
@bharathappali Ideally the profile should be able to cache most of the fields that make up the request for `createExperiment` API. This means that other than the experiment specific fields such as `kubernetes_objects`, `experiment_name` etc, almost everything else should be representable in the bulk profile

bharathappali:
Thanks for the review @dinogun 

I completely agree with your thought on cache all the required data which is needed to create a exp JSON via bulk, I have separated the options into two parts `cluster` & `recommendation`

So any datapoint in exp json (now / in future) will be falling into one of these options, i.e either it could be a cluster level data like list of namespaces, labels, cluster type, profiles etc or the recommendation setting like threshold, models, terms etc

So to avoid future DB schema changes I have opted these things to be in JSONB format so that we can resolve things at serialisation and deserialisation end with no changes in the schema end.

Please let me know your thoughts on this. Thanks in advance!

bharathappali:
One other field that I thought add was `experiment settings` but felt that it might be cluster specific so can be implemented in the cluster object at a later point in time

dinogun:
@bharathappali Please keep the same structure as we have in `createExperiment`, no need to invent a new structure here, only makes it confusing

migrations/lm/v111__create_kruize_bulk_profile_table.sql
```diff
@@ -0,0 +1,16 @@
+CREATE TABLE IF NOT EXISTS kruize_bulk_profile (
```
mbvreddy: As per my understanding, this table is used by optimizer to know the duration at which it has to fire bulk API calls. Today, it is only for single cluster, we are designing it to have support for mulitple cluster by single Kruize instance.

Firstly, I would suggest to name it as config instead of profile (already we have too many profiles and the content suits configuration than profile).
Secondly, we can think of naming it optimizer_config
Thirdly, we can maintain 1 configuration entry for each cluster so that we don't complicate the data model.

Optimizer get all active configurations of all cluster, and fire bulk call for each cluster as per configured schedule.

migrations/lm/v111__create_kruize_bulk_profile_table.sql
```diff
@@ -0,0 +1,16 @@
...
+    profile_name VARCHAR(255) PRIMARY KEY,
```
mbvreddy: Primary key can be `cluster_name` which is single value

migrations/lm/v111__create_kruize_bulk_profile_table.sql
```diff
@@ -0,0 +1,16 @@
...
+    clusters JSONB NOT NULL,
```
mbvreddy: Not required as it is superseded by cluster_name

migrations/lm/v111__create_kruize_bulk_profile_table.sql
```diff
@@ -0,0 +1,16 @@
...
+    datasources JSONB NOT NULL,
```
mbvreddy, line 4: As we are storing only list of datasource names, we can use native array TEXT[] instead of JSONB. For e.g., Database store value like {"datasource1", "datasource2"}

Image

migrations/lm/v111__create_kruize_bulk_profile_table.sql
```diff
@@ -0,0 +1,16 @@
...
+    namespaces JSONB NOT NULL,
```
mbvreddy, line 5: Here as well. use Native Array instead of JSONB if the value is only text elements.

migrations/lm/v111__create_kruize_bulk_profile_table.sql
```diff
@@ -0,0 +1,16 @@
...
+    labels JSONB,
+    experiment_types JSONB NOT NULL,
```
mbvreddy, line 7: Here as well! use Native Array instead of JSONB if the value is only text elements.

migrations/lm/v112__create_optimizer_bulk_config_table.sql
```diff
@@ -0,0 +1,16 @@
...
+    clusters JSONB NOT NULL,
+    datasources JSONB NOT NULL,
+    namespaces JSONB NOT NULL,
+    labels JSONB,
+    experiment_types JSONB NOT NULL,
+    metadata_profile VARCHAR(255),
+    performance_profile VARCHAR(255) NOT NULL,
+    trial_settings JSONB,
+    recommendation_settings JSONB NOT NULL,
+    webhook_url VARCHAR(500),
```
mbvreddy, line 12: What does it represent? bulk API endpoint?
If yes, wouldn't it be part of the Optimizer application configuration?

mbvreddy (changes requested):

migrations/lm/v112__create_optimizer_bulk_config_table.sql
bharathappali: > What does it represent? bulk API endpoint?

It represents the optimizer webhook endpoint where kruize tries to update the creator of bulk profile if the profile is changed [by others]

migrations/lm/v111__create_kruize_bulk_profile_table.sql
bharathappali: Thanks for the review @mbvreddy 

Updated the profile name as config name and made it `not null` and have made cluster name as the primary key

migrations/lm/v111__create_kruize_bulk_profile_table.sql
bharathappali: Updated it to `cluster_name`

migrations/lm/v111__create_kruize_bulk_profile_table.sql
bharathappali: Agreed! Thanks for pointing this.

Updated to `TEXT[]`

migrations/lm/v111__create_kruize_bulk_profile_table.sql
bharathappali: Done

migrations/lm/v111__create_kruize_bulk_profile_table.sql
bharathappali: Done

kusumachalasani (approved):
lgtm!

dinogun (approved):
The new structure is now consistent with the createExperiment JSON, LGTM!

PR #1982, opened 2026-06-23
PR #1: Add database schema, core models, and validation utility for cluster support

gulati-aakriti:
## Description

This PR adds the foundational infrastructure for cluster name support in Kruize, including database schema, JPA entities, data models, and validation utilities.

Fixes # (issue)

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Docs update
- [ ] Requires DB changes

## How has this been tested?

- [ ] New Test X
- [ ] Functional testsuite

**Test Configuration**

## Checklist :dart:

- [x] Comments added
- [ ] Documentation updated
- [ ] Tests added or updated

## Summary by Sourcery

Add foundational support for associating data sources with validated cluster names across the application and database.

New Features:
- Allow datasources to store and expose associated cluster names via the domain model and public API.
- Introduce centralized utilities for validating and parsing cluster names according to DNS-1123 rules.

Enhancements:
- Extend datasource information handling to include cluster metadata while preserving backward compatibility.

Chores:
- Add a database migration to persist cluster name associations on the kruize_datasources table.

### DIFF
--- a/migrations/lm/v111__add_cluster_to_datasources.sql
+++ b/migrations/lm/v111__add_cluster_to_datasources.sql
@@ -1,0 +1,8 @@
+ALTER TABLE kruize_datasources
+ADD COLUMN IF NOT EXISTS clusters JSONB;
+
+-- Add comment for documentation
+COMMENT ON COLUMN kruize_datasources.clusters IS 'JSON array of cluster names associated with this datasource, e.g. ["cluster-1","cluster-2"]';
+
+-- Note: No default value is set to maintain backward compatibility
+-- Existing datasources will have NULL clusters, which is handled gracefully in the application code
--- a/src/main/java/com/autotune/common/datasource/DataSourceInfo.java
+++ b/src/main/java/com/autotune/common/datasource/DataSourceInfo.java
@@ -155,11 +155,15 @@ public class DataSourceInfo {
     }
 
     /**
-     * Returns the list of cluster names associated with this datasource
-     * @return List of cluster names, empty list if no clusters defined
+     * Returns the list of cluster names associated with this datasource.
+     * The {@code clusters} field is final and set once at construction time;
+     * all callers only read the list, so returning the same reference is safe
+     * and avoids allocating a new list on every call.
+     *
+     * @return list of cluster names; empty list if no clusters were provided
      */
     public List<String> getClusters() {
-        return new ArrayList<>(clusters);
+        return clusters;
     }
 
     @Override
--- a/src/main/java/com/autotune/common/datasource/DataSourceInfo.java
+++ b/src/main/java/com/autotune/common/datasource/DataSourceInfo.java
@@ -17,6 +17,8 @@ package com.autotune.common.datasource;
 
 import java.net.MalformedURLException;
 import java.net.URL;
+import java.util.ArrayList;
+import java.util.List;
 
 import com.autotune.analyzer.utils.AnalyzerConstants;
 import com.autotune.common.auth.AuthenticationConfig;
@@ -41,10 +43,15 @@ public class DataSourceInfo {
     private final String namespace;
     private final URL url;
     private AuthenticationConfig authenticationConfig;
+    private final List<String> clusters;
 
     private static final org.slf4j.Logger LOGGER = LoggerFactory.getLogger(DataSourceInfo.class);
 
     public DataSourceInfo(String name, String provider, String serviceName, String namespace, URL url, AuthenticationConfig authConfig) {
+        this(name, provider, serviceName, namespace, url, authConfig, null);
+    }
+
+    public DataSourceInfo(String name, String provider, String serviceName, String namespace, URL url, AuthenticationConfig authConfig, List<String> clusters) {
         this.name = name;
         this.provider = provider;
         if (null == url) {
@@ -55,6 +62,7 @@ public class DataSourceInfo {
         this.serviceName = serviceName;
         this.namespace = namespace;
         this.authenticationConfig = authConfig;
+        this.clusters = (clusters != null) ? new ArrayList<>(clusters) : new ArrayList<>();
     }
 
     /**
@@ -146,6 +154,14 @@ public class DataSourceInfo {
         LOGGER.debug("Authentication details for datasource {} have been updated.", this.name);
     }
 
+    /**
+     * Returns the list of cluster names associated with this datasource
+     * @return List of cluster names, empty list if no clusters defined
+     */
+    public List<String> getClusters() {
+        return new ArrayList<>(clusters);
+    }
+
     @Override
     public String toString() {
         return "DataSourceInfo{" +
@@ -154,6 +170,7 @@ public class DataSourceInfo {
                 ", serviceName='" + serviceName + '\'' +
                 ", namespace='" + namespace + '\'' +
                 ", url=" + url +
+                ", clusters=" + clusters +
                 '}';
     }
 }
--- a/src/main/java/com/autotune/database/table/KruizeDataSourceEntry.java
+++ b/src/main/java/com/autotune/database/table/KruizeDataSourceEntry.java
@@ -16,7 +16,13 @@
 
 package com.autotune.database.table;
 
+import com.autotune.utils.Utils;
+import com.fasterxml.jackson.databind.JsonNode;
 import jakarta.persistence.*;
+import org.hibernate.annotations.JdbcTypeCode;
+import org.hibernate.type.SqlTypes;
+
+import java.util.List;
 
 /**
  * This is a Java class named KruizeDataSourceEntry annotated with JPA annotations.
@@ -42,6 +48,12 @@ public class KruizeDataSourceEntry {
     private String serviceName;
     private String namespace;
     private String url;
+
+    /** JSONB array of cluster names, e.g. {@code ["cluster-a","cluster-b"]}. */
+    @JdbcTypeCode(SqlTypes.JSON)
+    @Column(columnDefinition = "jsonb")
+    private JsonNode clusters;
+
     @ManyToOne(cascade = CascadeType.PERSIST) // Cascade PERSIST to auto-save authentication entry
     @JoinColumn(name = "authentication_id", nullable = false) // Foreign key column in the datasource table
     private KruizeAuthenticationEntry kruizeAuthenticationEntry;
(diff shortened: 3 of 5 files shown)


kusumachalasani (approved):
lgtm!

bharathappali (approved):
LGTM

dinogun:
The TEXT field is too fragile and also does not lend itself to cross referencing with any future cluster related tables. It is not good from a maintainability pov either. Suggest that we atleast convert this to a JSONB field that supports an arraylist of clusters to begin with.

src/main/java/com/autotune/common/datasource/DataSourceInfo.java
```diff
@@ -55,6 +62,7 @@ public DataSourceInfo(String name, String provider, String serviceName, String n
         this.serviceName = serviceName;
         this.namespace = namespace;
         this.authenticationConfig = authConfig;
+        this.clusters = (clusters != null) ? new ArrayList<>(clusters) : new ArrayList<>();
```
bharathappali, line 65: Thanks for raising the PR @gulati-aakriti 

Any reason why we are creating a new copy of the list instead of using the existing object?

src/main/java/com/autotune/common/datasource/DataSourceInfo.java
```diff
@@ -146,6 +154,14 @@ public void updateAuthConfig(AuthenticationConfig newAuthConfig) {
         LOGGER.debug("Authentication details for datasource {} have been updated.", this.name);
     }
 
+    /**
+     * Returns the list of cluster names associated with this datasource
+     * @return List of cluster names, empty list if no clusters defined
+     */
+    public List<String> getClusters() {
+        return new ArrayList<>(clusters);
```
bharathappali: Here as well. Do we need a mutable copy of cluster list returned for some reason? Like does the caller needs a fresh copy everytime they call `getClusters`? If yes, does the caller just reads the data or updates the structure? 

If it's a read operation, then we can carry forward the same object reference than creating a new list everytime.

@gulati-aakriti please let me know your thoughts on this. Thanks!

src/main/java/com/autotune/database/table/KruizeDataSourceEntry.java
```diff
@@ -101,4 +113,22 @@ public String getUrl() {
     public void setUrl(String url) {
         this.url = url;
     }
+
+    public JsonNode getClusters() {
+        return clusters;
+    }
+
+    public void setClusters(JsonNode clusters) {
+        this.clusters = clusters;
```
bharathappali, line 122: I see that the field `clusters` can be `nullable`, so will it be a good idea to set the clusters after a `null` check? 

Or do we need the clusters to set to null for some usecase? if yes, Would like to know about it.

src/main/java/com/autotune/utils/Utils.java
```diff
@@ -253,4 +256,46 @@ public static Timestamp getTimeStampFrom(String format, String date) {
...
+     *   <li>Cannot exceed {@value MAX_CLUSTER_NAME_LENGTH} characters</li>
+     * </ul>
+     * Invalid entries are logged and skipped; the method never returns {@code null}.
+     *
+     * @param clustersNode the JsonNode stored in the {@code clusters} JSONB column
+     * @return list of valid cluster-name strings, never {@code null}
+     */
+    public static List<String> parseClusterList(JsonNode clustersNode) {
+        if (clustersNode == null || !clustersNode.isArray()) {
+            return new ArrayList<>();
```
bharathappali, line 288: Will it be a good idea to log if cluster node is null or not an array before returning?

src/main/java/com/autotune/utils/Utils.java
```diff
@@ -253,4 +256,46 @@ public static Timestamp getTimeStampFrom(String format, String date) {
...
+        }
+        List<String> result = new ArrayList<>();
```
bharathappali: From the JVM standpoint, will it be good to set the result to `null` and only initialize a list if results exist?

I'm more inclined to have a null check over the caller of this function rather than sending an empty list, coz the empty objects do look empty but have a reference and carry object headers with them which is a cost on heap [might be minimal till the next minor GC occurs]. More than memory, the processing time taken to create an object is higher than setting it to `null`

@gulati-aakriti Please let me know your thoughts on this

src/main/java/com/autotune/utils/Utils.java
```diff
@@ -253,4 +256,46 @@ public static Timestamp getTimeStampFrom(String format, String date) {
...
+        for (JsonNode node : clustersNode) {
+            if (!node.isTextual()) {
```
bharathappali: Will there be a chance of `node` being `null`? I feel if such case exist, calling `isTextual` on top of a null object can cause a null pointer. I'm sure that the possibility of such an event is very low but would be a good practice to make sure the object is not null before accessing it's functions. 

Please let me know your thoughts on this

bharathappali (changes requested):
Thanks for raising this PR @gulati-aakriti, I appreciate your efforts.

Here are my thoughts on the changes which we might need. I'm happy to discuss if you think some of the changes are not required.

src/main/java/com/autotune/common/datasource/DataSourceInfo.java
gulati-aakriti: The defensive copy is intentional. DataSourceInfo is an immutable value object as all its fields are final. The copy in the constructor protects the class from being affected if the caller modifies their original list after construction.

src/main/java/com/autotune/common/datasource/DataSourceInfo.java
gulati-aakriti: fixing it

src/main/java/com/autotune/database/table/KruizeDataSourceEntry.java
gulati-aakriti: clusters maps to a nullable JSONB column in PostgreSQL. A null value is valid and moerover  existing datasources that were added before cluster support was introduced will have null in this column, and the code handles that gracefully in getClusterList() via Utils.parseClusterList().

src/main/java/com/autotune/utils/Utils.java
gulati-aakriti: definitely, will add

src/main/java/com/autotune/utils/Utils.java
gulati-aakriti: Returning null would push a null-check obligation onto every single call site and looking at [#PR2](https://github.com/kruize/autotune/pull/1983), getClusters() result is passed directly to setClusterList() and serialized in DataSourceInfoAdapter. Both callers already check !clusters.isEmpty() or just iterate safely over an empty list. Returning null to represent "no clusters" conflates absence-of-data with an error condition, which is a footgun - a forgotten null check causes a NullPointerException at runtime. 

And the heap cost of an empty ArrayList is ~40 bytes and is short-lived, which the GC handles efficiently.

src/main/java/com/autotune/utils/Utils.java
gulati-aakriti: Sure will fix it

src/main/java/com/autotune/common/datasource/DataSourceInfo.java
bharathappali: I agree with you on this, Thanks!

src/main/java/com/autotune/common/datasource/DataSourceInfo.java
bharathappali: Thank you

src/main/java/com/autotune/database/table/KruizeDataSourceEntry.java
bharathappali: I agree

src/main/java/com/autotune/utils/Utils.java
bharathappali: I'm a bit skeptic about it but would not want to have excessive `null` checks at callers if that's the case. Approving it!

src/main/java/com/autotune/utils/Utils.java
bharathappali: Thanks

bharathappali (approved):
LGTM

src/main/java/com/autotune/utils/Utils.java
```diff
@@ -253,4 +256,51 @@ public static Timestamp getTimeStampFrom(String format, String date) {
...
+     *   <li>Must be a non-null textual node</li>
+     *   <li>Cannot be blank after trimming</li>
+     *   <li>Cannot exceed {@value MAX_CLUSTER_NAME_LENGTH} characters</li>
+     * </ul>
+     * Invalid entries are logged and skipped; the method never returns {@code null}.
+     *
+     * @param clustersNode the JsonNode stored in the {@code clusters} JSONB column
+     * @return list of valid cluster-name strings, never {@code null}
+     */
+    public static List<String> parseClusterList(JsonNode clustersNode) {
```
kusumachalasani, line 281: Can we include the validation for duplicates as well ?

gulati-aakriti: Added

kusumachalasani (approved):
lgtm!

migrations/lm/v111__add_cluster_to_datasources.sql
```diff
@@ -0,0 +1,8 @@
+ALTER TABLE kruize_datasources
+ADD COLUMN IF NOT EXISTS clusters JSONB;
```
mbvreddy, line 2: As clusters is list of cluster names served by a given data source, defining it as JSONB is overkill. I recommend to use native array TEXT[].

Image

src/main/java/com/autotune/database/table/KruizeDataSourceEntry.java
```diff
@@ -101,4 +113,22 @@ public String getUrl() {
...
+
+    /**
+     * Returns the clusters JSONB column parsed into a {@code List<String>}.
+     * Delegates to {@link Utils#parseClusterList(JsonNode)}.
+     *
+     * @return list of cluster names; empty list if the column is null or malformed
+     */
+    public List<String> getClusterList() {
+        return Utils.parseClusterList(clusters);
+    }
```
mbvreddy, line 133: This might need a rework if datatype is changed to TEXT[] from JSONB

mbvreddy (changes requested):

dinogun:
Now that we are all agreed on the JSONB considering expected future changes including multi cluster support, approving this PR

dinogun (approved):
LGTM

